PR #3, opened 2016-07-20
add the init maven project and generated code

AdoHe:
@xiang90 I just noticed this project, and I think it's a better place to implement etcd v3 java client even though I add some init commits on my own repo. In this pr, I just add the init maven project and generated code, I think next I will implement the specific service client.


skyao:
Hello, Adohe, why you commit the generated java code to git? We need't do this, there are so many files.

skyao:
I pulled another request like you, but no generated java files committed.

And for the grpc version, I suggest to use the latest version 0.15.0.

xiang90:
Let's do not check in the generate code for now. I will merge in @skyao initial proto setup.

AdoHe:
That's ok.

PR #7, opened 2016-07-21
some basic code and test base 

skyao:
1. change java version to 1.7, not 1.8
2. rollback rpc package
3. add etcd client builder
4. setup integration test and unit test 

we can run a simple client to connect KV service now.


AdoHe:
IMO, no need to add `EtcdNativeClient`, we must wrap all grpc logical in client side, hide that from end users.  Also there are several things I am considering before implement this real client: 1. how to do client side load balancer, etcd v3 go client support simple load balancer, java should do this also. 2. sync VS async?, I would like to implement an asynchronous client. 3. how to divide services implementation into different sub-client. Now etcdv3 has six kinds of services, if we put all logical in one client, that would be a huge one. @skyao @xiang90

skyao:
@AdoHe: I add this native client just because we don't have normal one now.

Once we finish the normal one, we can remove this native one.

answers to your questions:

1. load balancer: I suggest to do this later, we need to bring out the first one without LB
2. seems we need bring both of them
3. in native client, I try to split the client to service according to GRPC services. I suggest we can continue to do like this, but review and redesign the services.

xiang90:
@skyao @AdoHe Yes, we should hide all gRPC logic. We (@sijie and me) are putting together the KV service interface now. Then we can start implement it.

Let's use the default gRPC round-robin lb for now.

src/main/java/com/coreos/jetcd/EtcdClientBuilder.java
```diff
@@ -1,4 +1,77 @@
...
+    /**
+     * create a new builder.
+     *
+     * @return new builder
+     */
+    public static EtcdClientBuilder newBuilder() {
+        return new EtcdClientBuilder();
+    }
+
+    private String etcdHost;
```
xiang90: let's take a list of endpoints? an endpoint is a string with host:port?

xiang90:
@skyao Can we port these implementation into the interface @sijie defined here: https://github.com/coreos/jetcd/blob/master/src/main/java/com/coreos/jetcd/EtcdKV.java#L21?

Thank you!

raoofm:
@skyao may be i missed something here, any reason for moving to 1.7 from 1.8?

xiang90:
@raoofm I was not sure if we should go with 1.7 or 1.8. @sijie suggested me that we should just go with 1.8. What do you think?

sijie:
Just FYI. https://java.com/en/download/faq/java_7.xml for java 7. 

Either (7 or 8) works for me. No strong preference here.

AdoHe:
Personally I would like to use Java 1.7.

xiang90:
@AdoHe Can I have your wechat account?

AdoHe:
@xiang90 my wechat account is 2584691618

skyao:
There are too many conflicts since sijie has commit another implementation of EtcdClient, so I will give up this poll request and redo some work based on code base of sijie.

raoofm:
@xiang90 @skyao @AdoHe 
As @sijie has pointed out "After April 2015, Oracle will no longer post updates of Java SE 7" and it is already 30 months since the release of java8 and we are using java8 for 2 years at our organization and didn't face any problems. To use the latest mostly has its benefits unless there is a strong case to not to use java8. 
@AdoHe i think we should go with java8?

raoofm:
Never mind I see an issue was closed and you guys decided to go ahead with java8

PR #11, opened 2016-07-22
Improve maven build

lburgazzoli:
I've improved the maven pom by
- define version in properties block
- use maven-bundle-plugin to add osgi mete-data
- upgrade dependencies
- remove unused dependencies


AdoHe:
@lburgazzoli maybe we should use jdk version 1.7 now.

lburgazzoli:
@AdoHe no problem, it was reflecting what was in the build before, I'll amend it

lburgazzoli:
@AdoHe fixed, I've also added a profile to disable doclint on java8

PR #12, opened 2016-07-22
rollback package name to etcdserverpb

skyao:
In etcd server, the service package is "etcdserverpb".

if we use jetcd in our proto file, we can't access the servies like KV service.

So we have to rollback to "etcdserverpb".


skyao:
Test and verified.

PR #13, opened 2016-07-22
add init kv service implementation

AdoHe:
@xiang90 @skyao @sijie I add some init kv services implementaion ptal.


AdoHe:
Also some thought on multi endpoints load balancing: Currently grpc use `DummyLoadBalancerFactory` as the default load balancer factory, the `DummyLoadBalancer` will always pick-first. Basically we need to implement custom **NameResolver** and **LoadBalancer** if we want to support multi endpoints.

skyao:
Seems there are some conflicts in class EtcdClientBuilder.

Issue #16, opened 2016-07-22
Client interfaces


lburgazzoli:
I'm wondering if it could make sense to enhance client interface with convenient methods i.e:
- Options should not be mandatory:

``` java
ListenableFuture<PutResponse> put(ByteString key, ByteString value);
```
- Leverage java 8 functional interfaces (if project will be moved to java 8 of course):

``` java
void put(
   ByteString key, ByteString value, 
   PutOption option, 
   Consumer<PutResponse> onResponse, Consumer<Throwable> onError);

void put(
   ByteString key, ByteString value, 
   Consumer<PutResponse> onResponse, Consumer<Throwable> onError);

void put(
   ByteString key, ByteString value, 
   Consumer<PutResponse> onResponse); throws an RuntimeException in case of error
```

jurmous:
My preference would be to return java 8 CompletableFuture on any of those put calls. It seems to be the route for futures/listeners in java going forward. You would then be able to do this:

```
future.handle((ok, ex) -> {
    if (ok != null) {
        // handle ok
    } else {
        handle error
    }
});
```
or just ``` future.get() ``` for sync call and throws ExecutionException

More:
http://www.nurkiewicz.com/2013/05/java-8-definitive-guide-to.html
http://download.java.net/lambda/b88/docs/api/java/util/concurrent/CompletableFuture.html

lburgazzoli:
The only problem is that grpc layer returns a ListenableFuture so we neeed to wrap/convert it, or is ther any better way ?

AdoHe:
SGTM.

xiang90:
@lburgazzoli we should do suggestion 1 for sure. can you help to improve that?

/cc @sijie what's your opinion on suggestion2?

lburgazzoli:
I've a patch (https://github.com/lburgazzoli/coreos-jetcd/commit/9f287c89a3bc0e22b3f782dff39f9c4b2e389e47) that should implement both (suggestion 2 using CompletableFutureas suggested by @jurmous). It includes an additional dependency at the moment for quick prototyping

jurmous:
Ah I did not look into grpc yet... Wrapping would create overhead but at the same time it is I think good to hide away implementation details which could change in the future. Don't know if there are alternatives with grpc and guava ListenableFutures than wrapping.

lburgazzoli:
Let me know what you think about lburgazzoli@9f287c8 so I can send a PR

sijie:
re the second question:

[...]

Issue #5, opened 2016-07-20
JDK version


skyao:
One question, which JDK version should we support in jetcd?

java8? or java 7?

skyao:
We decide to java7 now. Now java 8 limitation.

skyao:
Reopen this topic.

For me, java8 is enough. 

I don't know if there is some special requirement for java7.

jurmous:
Java7 is officially not supported anymore and Java 8 would allow to base the async calls on CompletableFuture. On HBase (where I also worked on Netty client besides etcd4j) they did not want to expose Netty futures so they decided to use CompletableFuture as the return. It seems that it is the most future ready way to do async calls. 

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html

AdoHe:
@skyao I think we need to make a decision about the jdk version from beginning.

skyao:
Java7 is officially not supported anymore.

This is my point to only support java8.

lburgazzoli:
+1 for java 8 only

I also think we should limit as much as possible to expose Guava's specific classes and eventually repackage it as Guava is sometime an issue if another incompatible version is already provided via dependency chain.

xiang90:
OK. Let's do java8 only. Thanks for your opinions.

PR #17, opened 2016-07-22
convenience methods and java version

lburgazzoli:
Includes code for:
- convenient method #16 
- java version #5


src/main/java/com/coreos/jetcd/EtcdHelper.java
```diff
@@ -0,0 +1,15 @@
+package com.coreos.jetcd;
```
xiang90: remove this file for now? nothing uses it.

xiang90:
other than that, lgtm

lburgazzoli:
done

xiang90:
thanks.

PR #20, opened 2016-07-24
initial txn impl

xiang90:
/cc @sijie @skyao @AdoHe @heyitsanthony @hongchaodeng


src/main/java/com/coreos/jetcd/op/Cmp.java
```diff
@@ -0,0 +1,65 @@
...
+                compareBuiler.setResult(Compare.CompareResult.EQUAL);
+                break;
+            case GREATER:
+                compareBuiler.setResult(Compare.CompareResult.GREATER);
+                break;
+            case LESS:
+                compareBuiler.setResult(Compare.CompareResult.LESS);
+                break;
+            default:
+                throw new IllegalArgumentException("Unexpected compare type");
```
hongchaodeng: I  think you should throw IllegalArgumentException in the constructor.

xiang90: ok. sure. i will make the change.

AdoHe:
After discussion with xiang, we decide to merge this now, I will add exception check later.

PR #21, opened 2016-07-24
test: run etcd before testing starts

xiang90:


scripts/run_etcd.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+mkdir -p external
+cd external
+
+git clone https://github.com/coreos/etcd
+cd etcd
+echo "Building etcd..."
+./build
```
heyitsanthony: should probably fetch the latest release?

xiang90: sure. i will merge this simple solution first. then i will do the git tricks to build the latest release.

PR #24, opened 2016-07-25
update EtcdKVImpl get&delete request builder

AdoHe:


src/main/java/com/coreos/jetcd/op/Op.java
```diff
@@ -113,5 +113,4 @@ RequestOp toRequestOp() {
             return RequestOp.newBuilder().setRequestDeleteRange(delete).build();
         }
     }
-
-}
+}
```
xiang90: add \n?

xiang90:
lgtm

src/main/java/com/coreos/jetcd/op/Op.java
AdoHe: just remove the extra "\n".

AdoHe:
@xiang90 I am considering whether we should re-use `PutOp` etc in `EtcdKVImpl.java` , now there are some duplicated code in `Op.java` and `EtcdKVImpl.java`.

PR #26, opened 2016-07-25
add abstract name resolver factory

AdoHe:
@lburgazzoli ptal.


AdoHe:
@lburgazzoli I just add init discovery service based name resolver, still wip. ptal

lburgazzoli:
Any chance to get it merged, maybe in a different branch ?

src/main/java/com/coreos/jetcd/resolver/DiscoveryEtcdNameResolver.java
```diff
@@ -0,0 +1,89 @@
...
+import java.net.URI;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+
+/**
+ * A discovery service based {@link NameResolver}. DiscoveryEtcdNameResolver will
+ * query configured discovery URL periodically to get endpoints addresses.
+ */
+public class DiscoveryEtcdNameResolver extends NameResolver {
```
xiang90: if you refer to public discovery.etcd.io, it does not really work. discovery url only expose peer url, not client url.

xiang90: i think the first naming resolver would be a dns srv resolver with client prefix. see https://github.com/coreos/etcd/blob/master/Documentation/op-guide/clustering.md#dns-discovery

xiang90:
@lburgazzoli @AdoHe I think we need to have srv discover first. the discover token one does not really work today. see my inline comments.

src/main/java/com/coreos/jetcd/resolver/DiscoveryEtcdNameResolver.java
AdoHe: Yes, I am wondering whether we should add a discovery service based name resolver.

AdoHe:
@xiang90 so perhaps we just need two name resolvers: `simple` and `dns+srv`.

xiang90:
@AdoHe What does simple do? nothing but return the original endpoints?

AdoHe:
@xiang90 Yes, simple name resolver just write the original endpoints to listener.

xiang90:
sure. thanks.

AdoHe:
@lburgazzoli I will update this today and merge it so you can start the srv based name resolver.

xiang90:
lgtm. can you please squash the commits into one?

PR #27, opened 2016-07-26
add Lease implement

StupidHod:
I have added a interface, implement and a simple test for Lease.


src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -14,15 +15,22 @@
     private final ManagedChannel channel;
 
     private final EtcdKV kvClient;
+    private final EtcdLease leaseClient;
 
     public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder builder) {
-        this.endpoints = (String[])builder.endpoints().toArray();
-        this.channelBuilder = channelBuilder != null ? channelBuilder : ManagedChannelBuilder.forAddress("localhost", 2379);
+        this.endpoints = new String[builder.endpoints().size()];
+        builder.endpoints().toArray(this.endpoints);
```
AdoHe: basically there is no need to construct a new string array.

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -14,15 +15,22 @@
...
+        this.channelBuilder = channelBuilder != null ? channelBuilder :
+                ManagedChannelBuilder
+                        .forAddress("localhost", 2379);
         this.channel = this.channelBuilder.build();
+
```
AdoHe: I would think one `\n` is enough.

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -33,4 +41,13 @@ public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder bui
     public EtcdKV newKVClient(KVGrpc.KVFutureStub stub) {
         return new EtcdKVImpl(stub);
     }
+
+    public EtcdLease newLeaseClient(ManagedChannel channel) {
+        return new EtcdLeaseImpl(channel);
+    }
+
+    public EtcdLease getLeaseClient(){
```
AdoHe: I am doubting whether we should add getter methods here. Can keep this for now.

src/main/java/com/coreos/jetcd/EtcdLease.java
```diff
@@ -0,0 +1,21 @@
...
+
+import com.coreos.jetcd.api.LeaseGrantResponse;
+import com.coreos.jetcd.api.LeaseKeepAliveResponse;
+import com.coreos.jetcd.api.LeaseRevokeResponse;
+import com.google.common.util.concurrent.ListenableFuture;
+
+/**
+ * Created by mmj on 16-7-25.
+ */
+public interface EtcdLease {
```
AdoHe: Please add some comments here to explain what method actually do.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,260 @@
...
+
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.*;
+
+/**
+ * Created by mmj on 16-7-25.
+ */
```
AdoHe: should better remove this comments.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,260 @@
...
+        /*
+        Init lease stub
+         */
+        this.channel = channel;
+        this.leaseFutureStub = LeaseGrpc.newFutureStub(this.channel);
+        this.leaseStub = LeaseGrpc.newStub(channel);
+
+
+        keepAlives = new ConcurrentHashMap<>();
+        lock = new Object();
```
AdoHe: `private final Object lock = new Object();` would be better.

AdoHe: There are more fields that you can initialize when you declare.

AdoHe: Also please pay attention to code format, thanks.

src/main/java/com/coreos/jetcd/EtcdClient.java
StupidHod: I encountered a  exception(java.lang.ClassCastException) here without this change in com.coreos.jetcd.EtcdLeaseTest  , =,=

src/main/java/com/coreos/jetcd/EtcdClient.java
StupidHod: Ok, got it.

src/main/java/com/coreos/jetcd/EtcdLease.java
StupidHod: Ok, tks, I will correct it.

AdoHe:
I would review `EtcdLeaseImpl` more carefully later, thanks @StupidHod for your great work.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: Ok, thanks for your kind review, I will correct it more carefully.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+    private StreamObserver<LeaseKeepAliveRequest> keepAliveRequestStreamObserver;
+
+    /**
+     * KeepAlive Response Streamer, receive keep alive response from this stream and update the
+     * nextKeepAliveTime and deadline of the leases.
+     */
+    private StreamObserver<LeaseKeepAliveResponse> keepAliveResponseStreamObserver;
+
+    public EtcdLeaseImpl(final ManagedChannel channel) {
+
```
xiang90: can we remove this extra empty line? this function is short.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+        this.channel = channel;
+        this.leaseFutureStub = LeaseGrpc.newFutureStub(this.channel);
+        this.leaseStub = LeaseGrpc.newStub(channel);
+
+        resetKeepAliveService();
+    }
+
+    /**
+     * Init the request stream to etcd
+     * start schedule to keep heartbeat to keep alive  and remove dead leases
```
xiang90: extra space between `keep alive` and `and`

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+     * @return
+     */
+    @Override
+    public ListenableFuture<LeaseGrantResponse> grant(long ttl) {
+        LeaseGrantRequest leaseGrantRequest =
+                LeaseGrantRequest.newBuilder()
+                        .setTTL(ttl)
+                        .build();
+        return leaseFutureStub.leaseGrant(leaseGrantRequest);
+
```
xiang90: remove extra empty line.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    /**
+     * keep alive one lease in background
+     * @param leaseId
+     */
+    @Override
+    public void keepAlive(long leaseId) {
+        synchronized (lock){
+            if(keepAlives.containsKey(leaseId)){
```
xiang90: if (!keepAlives.containsKey(leaseId)

xiang90: move the code block in the else statement into if statement.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+
+    /**
+     * keep alive one lease in background
+     * @param leaseId
+     */
+    @Override
+    public ListenableFuture<LeaseKeepAliveResponse> keepAliveOnce(long leaseId) {
+
+        /**
+         * to be completed, I will return a ListenableFuture value in the future
```
xiang90: throw unsupported exception for now.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+                leaseStub.leaseKeepAlive(keepAliveResponseStreamObserver);
+        requestObserver.onNext(newKeepAliveRequest(leaseId));
+        requestObserver.onCompleted();
+
+
+        return null;
+    }
+
+    /**
+     *Scan all the leases and seed keep alive request to etcd server
```
xiang90: * Scan (add empty space)

xiang90:
general approach looks good. @sijie @AdoHe @hongchaodeng @lburgazzoli or @skyao might provide more suggestions.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Implementation of lease client
+ */
+public class EtcdLeaseImpl implements EtcdLease {
+
+    private final static int DEFAULT_TTL = 5000;
+
+    private Object lock = new Object();;
```
sijie: Any idea why use a separate object rather than synchronizing on this? did I miss anything here?

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+
+    /**
+     * gRPC channel
+     */
+    private ManagedChannel channel;
```
sijie: final?

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+    private final LeaseGrpc.LeaseFutureStub leaseFutureStub;
+    private final LeaseGrpc.LeaseStub leaseStub;
+
+    /**
+     * Timer schedule to send keep alive request
+     */
+    private ScheduledExecutorService keepAliveSchedule;
```
sijie: keepAliveScheduler?

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+    private ScheduledExecutorService deadLineSchedule;
```
sijie: deadLineScheduler?

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+            @Override
+            public void onCompleted() {
+
+            }
+        };
+
+        /**
+         * Start heartbeat schedule to keep alive leases
+         */
+        keepAliveSchedule = Executors.newSingleThreadScheduledExecutor();
```
sijie: Can you explain more about the threading model here? For example, 

- why these are separated two schedulers?
- why they are single thread schedulers?
- if they are single thread scheduler and you also use concurrent hash map, why do you need additional 'Object' for synchronization?

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+
+        return null;
+    }
+
+    /**
+     *Scan all the leases and seed keep alive request to etcd server
+     */
+    private void keepAliveExecutor(){
+        synchronized (lock){
+            long now = System.currentTimeMillis();
```
sijie: Have you considered using nanos than millis? As the lease here is more about duration not absolute time, nanos will have better performance.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+            }
+        };
+
+        /**
+         * Start heartbeat schedule to keep alive leases
+         */
+        keepAliveSchedule = Executors.newSingleThreadScheduledExecutor();
+        keepAliveSchedule.scheduleAtFixedRate(
+                ()-> keepAliveExecutor()
+                , 0, 500, TimeUnit.MILLISECONDS);
```
sijie: It would be good to make the lease TTL configurable.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+    public void close(){
+        synchronized (lock){
+            if(keepAliveRequestStreamObserver!=null){
+                keepAliveRequestStreamObserver.onCompleted();
+                keepAliveRequestStreamObserver = null;
+            }
+        }
+    }
+
+    public  interface EtcdLeaseHandler{
```
sijie: Any idea why put an interface within an implementation class? it seems to be a bit strange to me.

src/test/java/com/coreos/jetcd/EtcdLeaseTest.java
```diff
@@ -0,0 +1,74 @@
...
+            }
+
+            @Override
+            public void onFailure(Throwable throwable) {
+                System.out.println(throwable);
+            }
+        });
+        try {
+
+            assert  (leaseId = responseListenableFuture.get().getID()) >0;
```
sijie: why not use assertions provided by the test library ?

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -14,15 +15,20 @@
     private final ManagedChannel channel;
 
     private final EtcdKV kvClient;
+    private final EtcdLease leaseClient;
 
     public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder builder) {
```
sijie: I probably missed bunch of changes here. But I feel it is a bit strange to have circle dependency between builder and client.

Ideally it would be:
- Builder builds a client
- Builder constructs a configuration that based on application's build methods
- Builder passes configuration to client rather than pass builder itself to the client.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: Ok, I will remove this, I want to do something here at begining.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: Got it, good skill, tks.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: - The deadLineScheduler periodically scan the leases and remove dead lease.
- The keepAliveScheduler periodically scan the leases and send request to keep alive the leases. 

I  think about to do them in one scheduler, but the go version do them in two goroutine so I follow, I think you are right we can it in one scheduler and cost less.

I use sync method because the `public void keepAlive(long leaseId) ` will call by user with different thread which will insert value to the map and with my origin design there are two thread schedulers.

You are right, I have overdone with sync, the concurrent hash map is enough and I will remove the lock object.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: As the scheduler run 500ms periodically, so I think millis maybe is enought.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: I will change this, tks.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: Ok, I will move it to EtcdLease interface.

src/test/java/com/coreos/jetcd/EtcdLeaseTest.java
StupidHod: I will replace it with assertions, tks for your kind review.

src/main/java/com/coreos/jetcd/EtcdLease.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.coreos.jetcd.api.LeaseGrantResponse;
+import com.coreos.jetcd.api.LeaseKeepAliveResponse;
+import com.coreos.jetcd.api.LeaseRevokeResponse;
+import com.google.common.util.concurrent.ListenableFuture;
+
+/**
+ * Interface of Lease talking to etcd
+ */
+public interface EtcdLease {
```
hongchaodeng: What about `Close()` method?

src/main/java/com/coreos/jetcd/EtcdLease.java
```diff
@@ -0,0 +1,40 @@
...
+     * @param leaseId the id of lease
+     * @return
+     */
+    ListenableFuture<LeaseRevokeResponse> revoke(long leaseId);
+
+    /**
+     * keep alive one lease in background
+     * @param leaseId
+     */
+    void keepAlive(long leaseId);
```
hongchaodeng: Compared with the Go definition:
```
KeepAlive(...) (<-chan *LeaseKeepAliveResponse, error)
```
How does user get the response of each heartbeat?

src/main/java/com/coreos/jetcd/EtcdLease.java
```diff
@@ -0,0 +1,40 @@
...
+ * Interface of Lease talking to etcd
+ */
+public interface EtcdLease {
+
+    /**
+     * New a lease with ttl value
+     * @param ttl ttl value
+     * @return
+     */
+    ListenableFuture<LeaseGrantResponse> grant(long ttl);
```
hongchaodeng: Why are we using `ListenableFuture` here instead of `Future`? Specifically, I am wondering in what case it will have callback?

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
hongchaodeng: deadlineScheduler :)

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+        void onLeaseExpired(long leaseId);
+        void onThrowable(Throwable throwable);
```
hongchaodeng: onError

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+                keepAliveRequestStreamObserver=null;
+            }
+            keepAliveRequestStreamObserver = leaseStub.leaseKeepAlive(keepAliveResponseStreamObserver);
+        }
+    }
+
+    /**
+     * close all the keep alive thread and process
+     */
+    public void close(){
```
hongchaodeng: Should we return Future for consistency with all other methods?

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+                }
+            }
+        }
+    }
+
+    /**
+     * This method update the deadline and NextKeepAlive time
+     * @param leaseKeepAliveResponse The response receive from etcd server
+     */
+    public void processKeepLiveRespond(LeaseKeepAliveResponse leaseKeepAliveResponse){
```
hongchaodeng: processKeepaliveResponse

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+        synchronized (lock){
+            if(keepAlives.containsKey(leaseKeepAliveResponse.getID())){
```
hongchaodeng: If it doesn't contain the key, is it something dangerous? Can we throw an exception? Or at least do some logging?

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+    /**
+     * keep alive one lease in background
+     * @param leaseId
+     */
+    @Override
+    public void keepAlive(long leaseId) {
+        synchronized (lock){
+            if(keepAlives.containsKey(leaseId)){
+
+            }else{
```
hongchaodeng: Format the code like:
```java
if (...) {
} else {
}
```

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+                Lease lease = new Lease(leaseId);
+                lease.setNextKeepAlive(System.currentTimeMillis());
```
hongchaodeng: Can you change it to:
```java
now := System.currentTimeMillis()
lease.SetNext(now).setDeadline(now + firstTimeout)...
```

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
hongchaodeng: It's better to keep these two tasks as one. Both will touch the shared lease data structure and it will require additional locking. IMHO, it would be better to keep it simple

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+
+        leaseHandler = etcdLeaseHandler;
+        /*
+        Init lease stub
+         */
+        this.channel = channel;
+        this.leaseFutureStub = LeaseGrpc.newFutureStub(this.channel);
+        this.leaseStub = LeaseGrpc.newStub(channel);
+
+        resetKeepAliveService();
```
hongchaodeng: Don't start any thread in constructor.
We should have a start() for it and return Future.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    /**
+     * Init the request stream to etcd
+     * start schedule to keep heartbeat to keep alive  and remove dead leases
+     */
+    public void resetKeepAliveService(){
+
+        newRequestStream();
```
hongchaodeng: `initRequestStream()`
It doesn't return anything here.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,300 @@
...
+
+    private LeaseKeepAliveRequest newKeepAliveRequest(long leaseId){
+        return LeaseKeepAliveRequest.newBuilder()
+                        .setID(leaseId)
+                        .build();
+    }
+
+    private void newRequestStream(){
+        synchronized (lock){
+            if(keepAliveRequestStreamObserver!=null){
```
hongchaodeng: In what case will it not be null? This hints that some routine are being rerun on error case, which makes me worried.

src/main/java/com/coreos/jetcd/EtcdLease.java
StupidHod: The ListenableFuture is the default return type of gRPC, so I use it here. #16 have talk about this issue.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: Ok, I will modfiy its name. I want to add a method named onResponse to give user response, what do you think about.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: Tks, got it.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: Got it.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: Good idea, tks.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: Ok, I have add one pair methods, what's about the Future reference?
`    /**
     * Init the request stream to etcd
     * start schedule to keep heartbeat to keep alive and remove dead leases
     */
    void startKeepAliveService() throws LeaseKeepAliveServiceAlreadyStart;

    /**
     * end the schedule for keep alive and remove dead leases
     */
    void closeKeepAliveService() throws LeaseKeepAliveServiceNotStart;`

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: Ok, got it.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: This may happened as the gRPC stream ended on error, the Lease client may restart the service this is what's I considered.

src/main/java/com/coreos/jetcd/EtcdClient.java
AdoHe: @sijie Perhaps we need something like `EtcdClientConfig`, `EtcdClientBuilder` is responsible for generate such config and pass it when construct a real `EtcdClient`.

src/main/java/com/coreos/jetcd/EtcdLease.java
AdoHe: Basically we need a handler, we can enable callbacks here.

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -0,0 +1,285 @@
...
+    public EtcdLeaseImpl(final ManagedChannel channel, EtcdLeaseHandler etcdLeaseHandler, long scanPeriod) {
+
+        this.leaseHandler = etcdLeaseHandler;
+
+        /*
+        Init lease stub
+         */
+        this.channel = channel;
+        this.leaseFutureStub = LeaseGrpc.newFutureStub(this.channel);
+        this.leaseStub = LeaseGrpc.newStub(channel);
```
AdoHe: `this.channel`

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
AdoHe: It depends on how we use this client. should user call `start` method directly? IMO start some background tasks in construct is not that bad.

src/main/java/com/coreos/jetcd/EtcdLease.java
StupidHod: Agree with you.

src/main/java/com/coreos/jetcd/EtcdLease.java
```diff
@@ -0,0 +1,77 @@
...
+        void onLeaseExpired(long leaseId);
+
+        /**
+         * onThrowable will be called when keep alive encountered exception
+         * @param throwable
+         */
+        void onThrowable(Throwable throwable);
+    }
+
+    class LeaseKeepAliveServiceNotStart extends Exception{}
```
AdoHe: Any advantages define these two classes inside interface? I can see reason we define `EtcdLeaseHandler` as inner interface, but for this I can not find good reason.

StupidHod: Ok, I define this as these Exception will only throw by this interface, should I move it to exception/ package?

Documentation/EtcdLease.md
```diff
@@ -0,0 +1,38 @@
+# Overview
+The EtcdLease implement interface `grant` and `revoke` to new a lease and revoke a lease to etcd.
+It can manage the leases to send keep alive request for leases.
+
+
+## `startKeepAliveService` function
```
xiang90: ## StartKeepAliveServe function

Remove the annotation? It looks strange. Try to click view to see the generated page.

Documentation/EtcdLease.md
```diff
@@ -0,0 +1,38 @@
...
+..* The function will start a backgroud scheduler `keepAliveSchedule` to send keep alive request for lease registered to EtcdLease Client.
```
xiang90: remove `..*`, use  `-` instead

Documentation/EtcdLease.md
```diff
@@ -0,0 +1,38 @@
...
+..* It will open a StreamObserver to etcd, send request to this StreamObserver can keep the lease alive
+..* It will new a ScheduledExecutorService to run `keepAliveExecutor` and `deadLineExecutor` periodically
+
+## `closeKeepAliveService` function
+..* This function will end the backgroud scheduler.
+..* It will close the StreamObserver to etcd
+
+## `keepAlive` function
+..* It will new a Lease object to keep the lease id and nextKeepAliveTime and DeadLine
```
xiang90: It will create a Lease object, and keep the lease alive.

Documentation/EtcdLease.md
```diff
@@ -0,0 +1,38 @@
...
+..* It will add the lease object to keepAlives map
```
xiang90: It will add the created lease object to the keepAlives map.

Documentation/EtcdLease.md
```diff
@@ -0,0 +1,38 @@
...
+..* the backgroud scheduler will send keep alive request for the registered leases.
```
xiang90: The background scheduler will send keep alive requests for the added lease when it approaches its nextKeepAliveTime.

Documentation/EtcdLease.md
```diff
@@ -0,0 +1,38 @@
...
+..* It will close the StreamObserver to etcd
+
+## `keepAlive` function
+..* It will new a Lease object to keep the lease id and nextKeepAliveTime and DeadLine
+..* It will add the lease object to keepAlives map
+..* the backgroud scheduler will send keep alive request for the registered leases. 
+
+## `keepAliveExecutor` function
+..* This function is called periodically by `keepAliveSchedule`
+..* It will scan the keepAlives map and get the leases that the nextKeepAliveTime before now
```
xiang90: It will scan the keepAlives map and find the leases that requires keepAlive requests to send based on its nextKeepAliveTime

Documentation/EtcdLease.md
```diff
@@ -0,0 +1,38 @@
...
+..* Send request to the StreamObserver for these leases
+
+## `deadLineExecutor` function
+..* This function is called periodically by `keepAliveSchedule`
+..* It will scan the keepAlives map and get the leases that the DeadLine before now
```
xiang90: It will scan the keepAlives map and find the leases that requires removed from the map based on its DeadLine

xiang90: Can we also mention why this can happen?

Documentation/EtcdLease.md
```diff
@@ -0,0 +1,38 @@
...
+..* call the on onLeaseExpired method for these leases
+
+## `keepAliveResponseStreamObserver` interface instance
+..* This StreamObserver is the stream that etcd send response
```
xiang90: The StreamObserver is the stream etcd uses to send responses.

Documentation/EtcdLease.md
```diff
@@ -0,0 +1,38 @@
...
+..* The gRPC will call onNext when the etcd send response
```
xiang90: The gRPC client calls onNext when etcd server delivers a response

Documentation/EtcdLease.md
```diff
@@ -0,0 +1,38 @@
...
+..* The onNext function will call `processKeepAliveRespond`
+
+## `processKeepAliveRespond` function
+..* It will get the lease instance
```
xiang90: It will get the lease instance based on the leaseID.

Documentation/EtcdLease.md
```diff
@@ -0,0 +1,38 @@
...
+..* if repond ttl < 0, then it removed the lease from keepAlives map and call `onLeaseExpired`
```
xiang90: If TTL < 0, it removes the lease from ...

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
```diff
@@ -77,7 +77,10 @@ public EtcdLeaseImpl(final ManagedChannel channel, EtcdLeaseHandler etcdLeaseHan
      * Init the request stream to etcd
      * start schedule to keep heartbeat to keep alive and remove dead leases
      */
-    public void startKeepAliveService(){
+    public void startKeepAliveService() throws LeaseKeepAliveServiceAlreadyStart{
```
xiang90: try to keep commit clean. for example, in this design doc commit, we should not change code.

hongchaodeng:
This PR is TOO LARGE.
@StupidHod Can you send three or more separate PRs instead:
- submit docs and interface design.
- submit real implementation.
- add lease functionality to client

src/main/java/com/coreos/jetcd/EtcdLeaseImpl.java
StupidHod: Tks, I will try to split my commits based on files.

Documentation/EtcdLease.md
StupidHod: Sorry for my poor document, tks for your kind review.

StupidHod:
@hongchaodeng Ok, I will try to rework with my PR.

StupidHod:
@AdoHe @xiang90 , I have recode the implement, look forward your review, tks.

xiang90:
closing in favor of #33.

xiang90:
@StupidHod Sorry. Actually can you create a new pull request for this? There are already too many comments.

StupidHod:
@xiang90 Ok.

PR #29, opened 2016-07-28
add etcd lease client interface

StupidHod:


src/main/java/com/coreos/jetcd/EtcdLease.java
```diff
@@ -0,0 +1,83 @@
...
+import com.google.common.util.concurrent.ListenableFuture;
+
+/**
+ * Interface of Lease talking to etcd
+ */
+public interface EtcdLease {
+
+    /**
+     * New a lease with ttl value
+     * @param ttl ttl value
```
xiang90: What is the unit? In seconds?

src/main/java/com/coreos/jetcd/EtcdLease.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param leaseId
+     * @return The keepAlive response
+     */
+    ListenableFuture<LeaseKeepAliveResponse> keepAliveOnce(long leaseId);
+
+    /**
+     * set EtcdLeaseHandler for EtcdLease
+     * @param etcdLeaseHandler
+     */
+    void setEtcdLeaseHandler(EtcdLeaseHandler etcdLeaseHandler);
```
xiang90: is it possible to set per lease handler instead of a global one?

src/main/java/com/coreos/jetcd/EtcdLease.java
StupidHod: Yes

xiang90: can we add this into doc?

src/main/java/com/coreos/jetcd/EtcdLease.java
StupidHod: OK. I  will add this feature.

src/main/java/com/coreos/jetcd/EtcdLease.java
StupidHod: Ok, I will add it with one commit.

StupidHod:
@xiang90 @AdoHe , I have added the support for per lease, hope for your review, tks.

src/main/java/com/coreos/jetcd/EtcdLease.java
```diff
@@ -0,0 +1,98 @@
...
+    ListenableFuture<LeaseKeepAliveResponse> keepAliveOnce(long leaseId);
+
+    /**
+     * set EtcdLeaseHandler for lease
+     *
+     * @param leaseId          id of the lease to set handler
+     * @param etcdLeaseHandler the handler for the lease
+     * @throws InvalidParameterException if lease do not exist
+     */
+    void setEtcdLeaseHandler(long leaseId, EtcdLeaseHandler etcdLeaseHandler) throws InvalidParameterException;
```
AdoHe: `InvalidParameterException` should be changed to other exception, if can not find one in Java, you can define custom one.

AdoHe:
LGTM, except the exception.

StupidHod:
@AdoHe Hi, I have add a NoSuchLeaseException to replace the InvalidParameterException.

AdoHe:
LGTM. @xiang90 any further comments here?

xiang90:
Lgtm

PR #30, opened 2016-07-29
add etcd kv service test

AdoHe:


AdoHe:
@xiang90 @hongchaodeng @skyao @sijie could someone help fix testng issue? I am not familiar with testng.

sijie:
I am not familiar with testng too. I am wondering why we didn't use junit?

AdoHe:
@skyao seems you add testng, could you help with this?

xiang90:
i think @StupidHod is trying testng. Can we make sure we only use one testing framework?

AdoHe:
@xiang90 I am ok about testng.

scripts/run_etcd.sh
```diff
@@ -1,12 +1,28 @@
 #!/bin/bash
 
-mkdir -p external
```
hongchaodeng: @AdoHe What do you think about separating this into another PR? We don't need to block this.

AdoHe: that's ok.

PR #31, opened 2016-07-29
add dns SRV based discovery

lburgazzoli:
I've added some basic code, let me know what should be amended, improved, etc

### DIFF
--- a/src/main/java/com/coreos/jetcd/resolver/AbstractEtcdNameResolver.java
+++ b/src/main/java/com/coreos/jetcd/resolver/AbstractEtcdNameResolver.java
@@ -35,7 +35,7 @@ public abstract class AbstractEtcdNameResolver extends NameResolver {
         this.executorResource = executorResource;
         URI nameUri = URI.create("//" + name);
         authority = Preconditions.checkNotNull(nameUri.getAuthority(),
-                "nameUri (%s) doesn't have an authority", nameUri);
+            "nameUri (%s) doesn't have an authority", nameUri);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


src/main/java/com/coreos/jetcd/resolver/AbstractEtcdNameResolver.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.annotation.concurrent.GuardedBy;
+
+import io.grpc.Attributes;
+import io.grpc.NameResolver;
+import io.grpc.ResolvedServerInfo;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+abstract class AbstractEtcdNameResolver extends NameResolver {
```
AdoHe: I believe we should reuse this `AbstractEtcdNameResolver` with `SimpleEtcdNameResolver`.

lburgazzoli: Once stabilized, yes definitively

src/main/java/com/coreos/jetcd/resolver/DnsSrvNameResolver.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @VisibleForTesting
+    DnsSrvNameResolver(String targetUri) {
+        super(URI.create(targetUri));
+        this.name = getServiceAuthority();
+    }
+
+    @Override
+    protected List<ResolvedServerInfo> resolve() {
+        try {
```
AdoHe: I think it's better to move logical here to a separate thread. Even in `SimpleEtcdNameResolver` I want to use a separate thread to update listener.

lburgazzoli: Should't be better to have such logic in AbstractEtcdNameResolver ?

lburgazzoli:
@AdoHe how does it look now ?

AdoHe:
LGTM. @lburgazzoli thanks very much.

Issue #25, opened 2016-07-25
dns discovery


lburgazzoli:
In @jurmous etcd4j's client, there were a simple dns srv discovery util to get a list of etcd instances, would it make sense to port it here ?

AdoHe:
@lburgazzoli thanks for pointing out this, I will take a look, if possible I will port it here, and actually we are going to implement one, you can see [here](https://github.com/coreos/jetcd/blob/master/src/main/java/com/coreos/jetcd/lb/EtcdNameResolverFactory.java). Thx :)

lburgazzoli:
@AdoHe best would be to have a common class to extend then I could el providing a dns imp

AdoHe:
@lburgazzoli grpc-java already has a dns based name resolver provider, I guess you are trying to implement a different one. So I would add an `AbstractEtcdNameResolverFactory` class, then you could provide a dns impl.

lburgazzoli:
I had a quick look at grpc's dns impl but it does not seems to use srv records. 
I would like to be able to leverage etcd or discovery options [1][2] like:

- dns:
```java
EtcdClientBuilder.newBuilder()
    .endpoints("dns+srv://_etcd-server._tcp.example.com")
    ...
```

- etcd: 
```java
EtcdClientBuilder.newBuilder()
    .endpoints("etcd://discovery.etcd.io/3e86b59982e49066c5d813af1c2e2579cbf573de")
    ...
```

If it make sese of course.


[1] https://coreos.com/etcd/docs/latest/clustering.html#etcd-discovery
[2] https://coreos.com/etcd/docs/latest/clustering.html#dns-discovery

AdoHe:
@lburgazzoli so I would suppose there are three ways to specify the endpoints:

1. dns:
```
EtcdClientBuilder.newBuilder()
    .endpoints("dns+srv://_etcd-server._tcp.example.com")
    ...
```
2. etcd discovery:
```
EtcdClientBuilder.newBuilder()
    .endpoints("etcd://discovery.etcd.io/3e86b59982e49066c5d813af1c2e2579cbf573de")
    ...
```
3. plain host:port list:
```
EtcdClientBuilder.newBuilder()
    .endpoints("http://10.0.0.1:2379", "http://10.0.0.2:2379")
    ...
```
Then we need to implement three types name resolver. wdyt?

lburgazzoli:
Could be, I do not yet know how the name resolver work but it is definitively be required for srv and etcd.

[...]

PR #32, opened 2016-07-30
add abstract etcd name resolver

AdoHe:
@lburgazzoli I just add `AbstractEtcdNameResolver` ptal. If any problem, we can discuss.

### DIFF
--- a/src/main/java/com/coreos/jetcd/resolver/AbstractEtcdNameResolver.java
+++ b/src/main/java/com/coreos/jetcd/resolver/AbstractEtcdNameResolver.java
@@ -1,0 +1,100 @@
+package com.coreos.jetcd.resolver;
+
+import com.google.common.base.Preconditions;
+import io.grpc.Attributes;
+import io.grpc.NameResolver;
+import io.grpc.ResolvedServerInfo;
+import io.grpc.internal.SharedResourceHolder;
+import io.grpc.internal.SharedResourceHolder.Resource;
+
+import javax.annotation.concurrent.GuardedBy;
+import java.net.URI;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+
+/**
+ * The abstract etcd name resolver, all other name resolvers should
+ * extend this one instead of {@link NameResolver}
+ */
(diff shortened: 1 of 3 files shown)


lburgazzoli:
I'm unsure about the multiple uri provided as constructor arg as [NameResolver.Factory::newNameResolver](http://www.grpc.io/grpc-java/javadoc/io/grpc/NameResolver.Factory.html#newNameResolver-java.net.URI-io.grpc.Attributes-) expects a single uri, am I wrong?

AdoHe:
Yes, `newNameResolver` expects a single target uri, which name resolver will resolve this target uri to specific addresses, I think in our case, the target uri would be: `dns+srv://_etcd-server._tcp.example.com` for dns based name resolver, and `etcd://etcd.v3.cluster` for simple name resolver.

lburgazzoli:
So I guess we should remove the multiple uri pram, isn't it ?

AdoHe:
Actually the reason I put multiple uri param here is users have to provide specific endpoints addresses when they use simple name resolver. For dns based name resolver, this info can be resolved through dns SRV records, for simple name resolver, this info must be configured manually.

AdoHe:
Maybe not a very good usage, if you have better suggestion, we can change this.

lburgazzoli:
I see the reason but I think abstract classes should stick to the interface contract/expected behavior, then simple resolver is free to change it.

AdoHe:
Yes, this sounds more reasonable, I will change in when I am near computer.

lburgazzoli:
Would it make sense, in the future, to contribute AbstractEtcdNameResolver and dns-srv discovery to o grpc-java ?

AdoHe:
@lburgazzoli I am not sure about this, maybe you can open an issue on grpc-java repo.

AdoHe:
@lburgazzoli I just update this, ptal. I am going to merge it if everything ok, thus you can reuse it.

lburgazzoli:
Fine with me, we can change it later on if needed

AdoHe:
@lburgazzoli I just merged this, please update PR #25

lburgazzoli:
@AdoHe done

PR #33, opened 2016-08-01
add auth interface and implemention

StupidHod:
add simple implementation for auth by convert gRPC interface

### DIFF
--- a/src/main/java/com/coreos/jetcd/EtcdAuth.java
+++ b/src/main/java/com/coreos/jetcd/EtcdAuth.java
@@ -1,0 +1,60 @@
+package com.coreos.jetcd;
+
+import com.coreos.jetcd.api.*;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.protobuf.ByteString;
+
+/**
+ * Interface of auth talking to etcd
+ */
+public interface EtcdAuth {
+
+    // ***************
+    // Auth Manage
+    // ***************
+
+    ListenableFuture<AuthEnableResponse> authEnable();
+
+    ListenableFuture<AuthDisableResponse> authDisable();
+
+    // ***************
+    // User Manage
+    // ***************
+
+    ListenableFuture<AuthUserAddResponse> userAdd(ByteString name, ByteString password);
+
+    ListenableFuture<AuthUserDeleteResponse> userDelete(ByteString name);
+
+    ListenableFuture<AuthUserChangePasswordResponse> userChangePassword(ByteString name, ByteString password);
+
+    ListenableFuture<AuthUserGetResponse> userGet(ByteString name);
+
+    ListenableFuture<AuthUserListResponse> userList();
+
+    // ***************
+    // User Role Manage
+    // ***************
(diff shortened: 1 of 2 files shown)


src/main/java/com/coreos/jetcd/EtcdAuthImpl.java
```diff
@@ -0,0 +1,173 @@
...
+ * Implementation of etcd auth client
+ */
+public class EtcdAuthImpl implements EtcdAuth{
+    
+    private ManagedChannel channel;
+    private AuthGrpc.AuthFutureStub authStub;
+    
+    
+    public EtcdAuthImpl(ManagedChannel channel){
+        this.channel=channel;
```
AdoHe: could we directly pass stub here?

AdoHe:
Generally LGTM. We may need to consider how to support connect to TLS secured etcd cluster or even support basic auth.

src/main/java/com/coreos/jetcd/EtcdAuthImpl.java
StupidHod: @AdoHe Tks, I will construct it directly by stub.

AdoHe:
Also please help add auth tests once we resolve testng issue @StupidHod .

StupidHod:
@AdoHe Ok, I have add some test in my local repositorie, wait for the resolution.

src/main/java/com/coreos/jetcd/EtcdAuth.java
```diff
@@ -0,0 +1,68 @@
...
+
+    ListenableFuture<AuthEnableResponse> authEnable();
+
+    ListenableFuture<AuthDisableResponse> authDisable();
+
+    // ***************
+    // User Manage
+    // ***************
+
+    ListenableFuture<AuthUserAddResponse> userAdd(ByteString name, ByteString passWord);
```
xiang90: password instead of passWord

src/main/java/com/coreos/jetcd/EtcdAuth.java
```diff
@@ -0,0 +1,68 @@
...
+    ListenableFuture<AuthRoleDeleteResponse> roleDelete(ByteString role);
+
+
+    /**
+     * Authentication with user name and password, the service will return token if authentication succeeds
+     * @param name
+     * @param passWord
+     * @return AuthenticateResponse with token or error information
+     */
+    ListenableFuture<AuthenticateResponse> authenticate(ByteString name, ByteString passWord);
```
xiang90: let's remove authenticate for now. Ideally, we should call authenticate internally inside the client library and setup authenticated connection for users. Users do not need to do this themselves. Or they have to attach the token to each requests, which is tedious.

StupidHod:
@xiang90 Tks, I have correct the word error and remove the authenticate function as your advice.

xiang90:
lgtm. Squash commits into one please.

xiang90:
@StupidHod Thanks a lot!

PR #35, opened 2016-08-04
add test for auth and merge auth to etcd client

StupidHod:
Look forward your review, @xiang90 @AdoHe .

### DIFF
--- a/scripts/run_etcd.sh
+++ b/scripts/run_etcd.sh
@@ -3,10 +3,8 @@
 mkdir -p external
 cd external
 
-git clone https://github.com/coreos/etcd
-cd etcd
-echo "Building etcd..."
-./build
-echo "Done!"
+curl -L https://github.com/coreos/etcd/releases/download/v3.0.4/etcd-v3.0.4-linux-amd64.tar.gz -o etcd-v3.0.4-linux-amd64.tar.gz
+tar xzvf etcd-v3.0.4-linux-amd64.tar.gz && cd etcd-v3.0.4-linux-amd64
+./etcd --version
 
-nohup bash -c 'bin/etcd &'
+nohup bash -c './etcd &'
--- a/src/main/java/com/coreos/jetcd/EtcdClient.java
+++ b/src/main/java/com/coreos/jetcd/EtcdClient.java
@@ -1,36 +1,148 @@
 package com.coreos.jetcd;
 
+import com.coreos.jetcd.api.AuthGrpc;
+import com.coreos.jetcd.api.AuthenticateRequest;
+import com.coreos.jetcd.api.AuthenticateResponse;
 import com.coreos.jetcd.api.KVGrpc;
+import com.coreos.jetcd.exception.AuthFailedException;
+import com.coreos.jetcd.exception.ConnectException;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.protobuf.ByteString;
+import io.grpc.CallCredentials;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
+import io.grpc.Metadata;
+import io.grpc.stub.AbstractStub;
+
+import java.util.concurrent.ExecutionException;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
  * Etcd Client
  */
 public class EtcdClient {
 
+    private static final String TOKEN = "token";
+
     private final ManagedChannelBuilder<?> channelBuilder;
     private final String[] endpoints;
     private final ManagedChannel channel;
 
     private final EtcdKV kvClient;
+    private final EtcdAuth authClient;
 
-    public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder builder) {
-        this.endpoints = (String[])builder.endpoints().toArray();
-        this.channelBuilder = channelBuilder != null ? channelBuilder : ManagedChannelBuilder.forAddress("localhost", 2379);
+    private KVGrpc.KVFutureStub kvStub;
+    private AuthGrpc.AuthFutureStub authStub;
+
+    public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder builder) throws ConnectException, AuthFailedException {
+        this.endpoints = new String[builder.endpoints().size()];
+        builder.endpoints().toArray(this.endpoints);
+        this.channelBuilder = channelBuilder != null ? channelBuilder : ManagedChannelBuilder.forAddress("localhost", 2379).usePlaintext(true);
+
         this.channel = this.channelBuilder.build();
 
-        KVGrpc.KVFutureStub kvStub = KVGrpc.newFutureStub(this.channel);
+        this.kvStub = KVGrpc.newFutureStub(this.channel);
+        this.authStub = AuthGrpc.newFutureStub(this.channel);
+
+        String token = getToken(builder);
+
+        if (token != null) {
+            this.authStub = setTokenForStub(authStub, token);
+            this.kvStub = setTokenForStub(kvStub, token);
+        }
 
         this.kvClient = newKVClient(kvStub);
+        this.authClient = newAuthClient(authStub);
     }
 
-        /**
-         * create a new KV client.
-         *
-         * @return new KV client
-         */
+    /**
+     * create a new KV client.
+     *
+     * @return new KV client
+     */
     public EtcdKV newKVClient(KVGrpc.KVFutureStub stub) {
         return new EtcdKVImpl(stub);
     }
+
+    private EtcdAuth newAuthClient(AuthGrpc.AuthFutureStub stub) {
+        return new EtcdAuthImpl(stub);
+    }
+
+    protected EtcdAuth getAuthClient() {
+        return authClient;
+    }
+
+    protected EtcdKV getKVClient() {
+        return kvClient;
+    }
+
+    /**
+     * add token to channel's head
+     *
+     * @param stub  the stub to attach head
+     * @param token the token for auth
+     * @param <T>   the type of stub
+     * @return the attached stub
+     */
+    private <T extends AbstractStub<?>> T setTokenForStub(T stub, String token) {
+        Metadata metadata = new Metadata();
+        Metadata.Key<String> TOKEN_KEY = Metadata.Key.of(TOKEN, Metadata.ASCII_STRING_MARSHALLER);
+        metadata.put(TOKEN_KEY, token);
+        CallCredentials callCredentials = (methodDescriptor, attributes, executor, metadataApplier) -> metadataApplier.apply(metadata);
+        return ((T) stub.withCallCredentials(callCredentials));
+    }
+
+    /**
+     * get token from etcd with name and password
+     *
+     * @param channel  channel to etcd
+     * @param name     auth name
+     * @param password auth password
+     * @return authResp
+     */
+    private ListenableFuture<AuthenticateResponse> authenticate(ManagedChannel channel, ByteString name, ByteString password) {
+
+        ListenableFuture<AuthenticateResponse> authResp = AuthGrpc.newFutureStub(channel)
+                .authenticate(AuthenticateRequest.newBuilder()
+                        .setNameBytes(name)
+                        .setPasswordBytes(password)
+                        .build());
+        return authResp;
+    }
+
+    /**
+     * get token with EtcdClientBuilder
+     *
+     * @param builder
+     * @return the auth token
+     * @throws ConnectException    This may be caused as network reason, wrong address
+     * @throws AuthFailedException This may be caused as wrong username or password
+     */
+    private String getToken(EtcdClientBuilder builder) throws ConnectException, AuthFailedException {
+
+        if (builder.getName() != null || builder.getPassword() != null) {
+
+            checkNotNull(builder.getName(), "username can not be null.");
+            checkNotNull(builder.getPassword(), "password can not be null.");
+            checkArgument(builder.getName().toStringUtf8().trim().length() != 0, "username can not be null.");
+            checkArgument(builder.getPassword().toStringUtf8().trim().length() != 0, "password can not be null.");
+
+            try {
+                return authenticate(this.channel, builder.getName(), builder.getPassword()).get().getToken();
+            } catch (InterruptedException ite) {
+                throw new ConnectException("connect to etcd failed", ite);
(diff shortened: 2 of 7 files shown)


src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -33,4 +67,46 @@ public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder bui
...
+
+    /**
+     * get token from etcd with name and password
+     *
+     * @param channel channel to etcd
+     * @param name auth name
+     * @param password auth password
+     * @return token string
+     */
+    private String getToken(ManagedChannel channel, ByteString name, ByteString password) {
```
xiang90: authenticate instead of getToken?

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -33,4 +67,46 @@ public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder bui
...
+        try {
+            ListenableFuture<AuthenticateResponse> authResp = AuthGrpc.newFutureStub(channel)
+                    .authenticate(
+                            AuthenticateRequest.newBuilder()
+                                    .setNameBytes(name)
+                                    .setPasswordBytes(password)
+                                    .build());
+            return authResp.get().getToken();
+        }catch (InterruptedException ie){
+            return null;
```
xiang90: we probably should throw out this exception. user should know the auth fails.

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -33,4 +67,46 @@ public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder bui
...
+    /**
+     * get token from etcd with name and password
+     *
+     * @param channel channel to etcd
+     * @param name auth name
+     * @param password auth password
+     * @return token string
+     */
+    private String getToken(ManagedChannel channel, ByteString name, ByteString password) {
+
```
xiang90: remove the empty line

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -14,15 +23,40 @@
...
         this.channel = this.channelBuilder.build();
 
-        KVGrpc.KVFutureStub kvStub = KVGrpc.newFutureStub(this.channel);
+        kvStub = KVGrpc.newFutureStub(this.channel);
+        authStub = AuthGrpc.newFutureStub(this.channel);
+
+        String token=null;
+
+        if(builder.getName()!=null && builder.getPassword()!=null){
```
xiang90: throw exception if one of them is empty and one of them is not.

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -14,15 +23,40 @@
...
+            token = getToken(this.channel, builder.getName(), builder.getPassword());
+        }
+
+        if(token!=null) {
+            /**
+             * add token to channel header
+             */
+            Metadata metadata = new Metadata();
+            Metadata.Key<String> TOKEN_KEY= Metadata.Key.of("token", Metadata.ASCII_STRING_MARSHALLER);
```
xiang90: in go, there is a utility func to set creds: https://github.com/coreos/etcd/blob/edcfcae332679cf02396d8a0f743bed047a06768/clientv3/client.go#L202

Is there a similar thing in java?

src/main/java/com/coreos/jetcd/EtcdClient.java
StupidHod: OK.

src/main/java/com/coreos/jetcd/EtcdClient.java
StupidHod: Hi, should we pack this exception with AuthFailedException, or just throw the origin exception?

src/main/java/com/coreos/jetcd/EtcdClient.java
StupidHod: Ok.

src/main/java/com/coreos/jetcd/EtcdClient.java
AdoHe: what we expect user to do if we throw this exception?

xiang90: password might be wrong. user should fill in the correct one. connection might be bad, need to retry.

src/main/java/com/coreos/jetcd/EtcdClient.java
AdoHe: @StupidHod you may take a look at `withCallCredentials` method of stub.

StupidHod: Hi, MetadataUtils.attachHeaders is a tool method to add metadata, which is short and easy to use.
MetadataUtils.attachHeaders has do the same thing as withCallCredentials.
`            authStub.withCallCredentials(new CallCredentials() {
                @Override
                public void applyRequestMetadata(MethodDescriptor<?, ?> methodDescriptor, Attributes attributes, Executor executor, MetadataApplier metadataApplier) {
                    metadataApplier.apply(metadata);
                }
            });`

xiang90: let's use withCallCredential. That is more "official" way to do this sort of thing

mitake:
@xiang90 sure, could you wait until this weekend? For my private reasons I cannot work on it immediately.

xiang90:
@mitake sure thanks.

src/test/java/com/coreos/jetcd/EtcdAuthClientTest.java
```diff
@@ -0,0 +1,199 @@
...
+     * @throws InterruptedException
+     */
+    @BeforeTest
+    public void setupEnv() throws AuthFailedException, ConnectException {
+        test = new Assertion();
+        etcdClient =
+                EtcdClientBuilder.newBuilder().endpoints("localhost:2379").build();
+        kvClient = etcdClient.getKVClient();
+        authClient = etcdClient.getAuthClient();
+
```
xiang90: remove empty line.

src/test/java/com/coreos/jetcd/EtcdAuthClientTest.java
```diff
@@ -0,0 +1,199 @@
...
+     * @throws ExecutionException
+     * @throws InterruptedException
+     */
+    @Test(dependsOnMethods = "testEnableAuth", groups = "authEnable")
+    public void setupAuthClient() throws AuthFailedException, ConnectException {
+        etcdClient =
+                EtcdClientBuilder.newBuilder().endpoints("localhost:2379").setName(userName).setPassword(password).build();
+        kvClient = etcdClient.getKVClient();
+        authClient = etcdClient.getAuthClient();
+
```
xiang90: remove empty line.

xiang90:
@StupidHod Can you fix the CI?

src/test/java/com/coreos/jetcd/EtcdAuthClientTest.java
xiang90: remove empty line.

xiang90:
@StupidHod Have you tested the auth path? Enable auth, add a role A with rw to key foo, add a user A, assign role A to User A. Without user A and its password, client cannot write to foo. With user A and its correct password, client can write to foo.

mitake:
@xiang90 @StupidHod I'm looking around this PR and want to try it. Does jetcd have an working example?

StupidHod:
@xiang90 Ok, I will try to fix the CI and add more tests.

StupidHod:
@mitake You can clone from my repository and switch to auth branch to run this test.

scripts/run_etcd.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
-
+git clone https://go.googlesource.com/go
+cd go
+git checkout master
+cd src
+./all.bash
+GOROOT_BOOTSTRAP=$GOROOT
```
hongchaodeng: Can we download etcd release instead of building it everytime?

AdoHe: @StupidHod @hongchaodeng actually in PR #30 I have update this, ptal.

StupidHod: @hongchaodeng You advice is good.

StupidHod: @AdoHe Your version seems better, you can overwrite my version latter, as I want to pass the test. =,=

hongchaodeng:
Hi @StupidHod , Can you try to use an IDE that formats java files :)
It would be much easier to read code like that. For example, the if-else block should have more spaces.

StupidHod:
@hongchaodeng OK, tks for your advice.

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -1,28 +1,70 @@
...
+
+        String token=null;
+
+        if(builder.getName()!=null && builder.getPassword()!=null){
+            token = authenticate(this.channel, builder.getName(), builder.getPassword());
+        }else if(!(builder.getName()==null&&builder.getPassword()==null))
+        {
+            throw new NullPointerException("Username or password can not be null.");
+        }
+        if(token!=null) {
```
hongchaodeng: Can you change the code path to:
```
if case1:
  throw e1
else if case2:
  throw e2

token = authenticate(...)
// use token below...
```

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -1,28 +1,70 @@
...
+            /**
+             * add token to channel header
+             */
+            Metadata metadata = new Metadata();
+            Metadata.Key<String> TOKEN_KEY= Metadata.Key.of(TOKEN, Metadata.ASCII_STRING_MARSHALLER);
```
hongchaodeng: This place has too many details and is confusing..
Can you help make them a function with a comprehensive function name?

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -33,4 +75,47 @@ public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder bui
...
+     * @param name auth name
+     * @param password auth password
+     * @return token string
+     * @throws  ConnectException This may be caused as network reason, wrong address
+     * @throws  AuthFailedException This may be caused as wrong username or password
+     */
+    private String authenticate(ManagedChannel channel, ByteString name, ByteString password) throws ConnectException, AuthFailedException {
+        try {
+            ListenableFuture<AuthenticateResponse> authResp = AuthGrpc.newFutureStub(channel)
+                    .authenticate(
```
hongchaodeng: move `authenticate` to previous line

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -33,4 +75,47 @@ public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder bui
...
+     * get token from etcd with name and password
+     *
+     * @param channel channel to etcd
+     * @param name auth name
+     * @param password auth password
+     * @return token string
+     * @throws  ConnectException This may be caused as network reason, wrong address
+     * @throws  AuthFailedException This may be caused as wrong username or password
+     */
+    private String authenticate(ManagedChannel channel, ByteString name, ByteString password) throws ConnectException, AuthFailedException {
```
hongchaodeng: Can we split this function into two? One is authenticate returning auth response; the other is get token (maybe we don't even need second func).

src/main/java/com/coreos/jetcd/EtcdClientBuilder.java
```diff
@@ -54,14 +59,48 @@ public EtcdClientBuilder endpoints(String ... endpoints) {
...
+    }
+
+    /**
+     * config etcd auth name
+     *
+     * @param name etcd auth name
+     * @return this builder
+     * @throws NullPointerException if name is null
+     */
+    public EtcdClientBuilder setName(ByteString name) {
```
hongchaodeng: why we are using ByteString?

src/main/java/com/coreos/jetcd/EtcdClient.java
StupidHod: OK.

src/main/java/com/coreos/jetcd/EtcdClientBuilder.java
StupidHod: @hongchaodeng , We use this as this is the string used by gRPC which support many encoding.

mitake:
@StupidHod I mean I want some examples that can be executed with running etcd servers. Or the test can be executed? Sorry, I don't know well about testing stuff of Java...

scripts/run_etcd.sh
```diff
@@ -3,10 +3,8 @@
...
-cd etcd
-echo "Building etcd..."
-./build
-echo "Done!"
+curl -L https://github.com/coreos/etcd/releases/download/v3.0.4/etcd-v3.0.4-linux-amd64.tar.gz -o etcd-v3.0.4-linux-amd64.tar.gz
+tar xzvf etcd-v3.0.4-linux-amd64.tar.gz && cd etcd-v3.0.4-linux-amd64
+./etcd --version
 
-nohup bash -c 'bin/etcd &'
+nohup bash -c './etcd &'
```
hongchaodeng: have a new line

src/test/java/com/coreos/jetcd/EtcdClientBuilderTest.java
```diff
@@ -33,7 +35,7 @@ public void testEndPoints_Verify_SomeEmpty() {
     }
 
     @Test(expectedExceptions = IllegalStateException.class)
-    public void testBuild_WithoutEndpoints() {
+    public void testBuild_WithoutEndpoints() throws AuthFailedException, ConnectException {
         builder.build();
     }
 }
```
hongchaodeng: have a new line at the end of the file

hongchaodeng:
@mitake 
I think we (probably @StupidHod ) can make some examples separately. Let's move forward with this PR.
LGTM after fixing the nits.

StupidHod:
@hongchaodeng Well, I will make some examples late, sorry for the nits issues, I have fixt them.

hongchaodeng:
LGTM.
Thanks @StupidHod for the work.

mitake:
@hongchaodeng @StupidHod I see, I'd like to check the behavior of the code with an working example later.

PR #36, opened 2016-08-05
add etcd config

linsimiao:
First try to commit EtcdConfig, hope there is some reviews, thanks!

### DIFF
--- a/src/main/java/com/coreos/jetcd/EtcdClient.java
+++ b/src/main/java/com/coreos/jetcd/EtcdClient.java
@@ -3,8 +3,6 @@ package com.coreos.jetcd;
 import com.coreos.jetcd.api.*;
 import com.coreos.jetcd.exception.AuthFailedException;
 import com.coreos.jetcd.exception.ConnectException;
-import com.coreos.jetcd.resolver.AbstractEtcdNameResolverFactory;
-import com.coreos.jetcd.resolver.SimpleEtcdNameResolverFactory;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.protobuf.ByteString;
 import io.grpc.CallCredentials;
@@ -13,8 +11,6 @@ import io.grpc.ManagedChannelBuilder;
 import io.grpc.Metadata;
 import io.grpc.stub.AbstractStub;
 
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.util.concurrent.ExecutionException;
 
 import static com.google.common.base.Preconditions.checkArgument;
@@ -28,9 +24,8 @@ public class EtcdClient {
     private static final String TOKEN = "token";
 
     private final ManagedChannelBuilder<?> channelBuilder;
-    private final String[] endpoints;
+    private EtcdConfig etcdConfig;
     private final ManagedChannel channel;
-    private final AbstractEtcdNameResolverFactory nameResolverFactory;
 
     private final EtcdKV kvClient;
     private final EtcdAuth authClient;
@@ -40,18 +35,12 @@ public class EtcdClient {
     private KVGrpc.KVFutureStub kvStub;
     private AuthGrpc.AuthFutureStub authStub;
 
-    public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder builder) throws ConnectException, AuthFailedException {
-        if (builder.getNameResolverFactory() != null) {
-            endpoints = null;
-            this.nameResolverFactory = builder.getNameResolverFactory();
-        } else {
-            //If no nameResolverFactory was set, use SimpleEtcdNameResolver
-            this.endpoints = new String[builder.endpoints().size()];
-            builder.endpoints().toArray(this.endpoints);
-            this.nameResolverFactory = getSimpleNameResolveFactory(this.endpoints);
-        }
-
-        this.channelBuilder = channelBuilder != null ? channelBuilder : ManagedChannelBuilder.forTarget("etcd").nameResolverFactory(nameResolverFactory).usePlaintext(true);
+    /**
+     * construct an EtcdClient with EtcdConfig
+     */
+    public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdConfig config) throws ConnectException, AuthFailedException {
+        this.etcdConfig = config;
+        this.channelBuilder = channelBuilder != null ? channelBuilder : ManagedChannelBuilder.forTarget("etcd").nameResolverFactory(config.nameResolverFactory).usePlaintext(true);
         this.channel = this.channelBuilder.build();
 
         this.kvStub = KVGrpc.newFutureStub(this.channel);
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -65,7 +65,6 @@
             <artifactId>slf4j-api</artifactId>
             <version>${slf4j.version}</version>
         </dependency>
-
```
AdoHe: no need to remove this empty line.

src/main/java/com/coreos/jetcd/EtcdConfig.java
```diff
@@ -0,0 +1,146 @@
...
+import java.security.cert.CertificateException;
+import java.util.Map;
+import java.util.List;
+import java.util.stream.Stream;
+
+
+public class EtcdConfig {
+
+    // endPoints is a list of URLs
+    private List<String> endPoints;
```
AdoHe: use `endpoints`.

src/main/java/com/coreos/jetcd/EtcdConfig.java
```diff
@@ -0,0 +1,146 @@
...
+    private String username;
+
+    // password is a password for authentication
+    private String password;
+
+    public List<String> getEndPoints() {
+        return endPoints;
+    }
+
+    public void setEndPoints(List<String> endPoints) {
```
AdoHe: the same.

src/main/java/com/coreos/jetcd/EtcdConfig.java
```diff
@@ -0,0 +1,146 @@
...
+        cfg.setCertfile((String) yamlMap.get("certfile"));
+        cfg.setKeyfile((String) yamlMap.get("keyfile"));
+        cfg.setcAfile((String) yamlMap.get("cAfile"));
+        cfg.setUsername((String)yamlMap.get("username"));
+        cfg.setPassword((String)yamlMap.get("password"));
+        if((boolean)yamlMap.get("insecureTransport") == true){
+            cfg.setSslContext(null);
+            return cfg;
+        }
+        else{
```
AdoHe: please pay attention to the code format.

AdoHe: I am not sure whether we should support yaml in Java.

src/main/java/com/coreos/jetcd/EtcdConfig.java
```diff
@@ -0,0 +1,146 @@
...
+
+    //keyfile is the filename of the key file.
+    private String keyfile;
+
+    //cAfile is the filename fo the CA file.
+    private String cAfile;
+
+    // sslContext holds the client secure credentials, if any.
+    private SslContext sslContext;
+
```
AdoHe: Actually there is a `GrpcSslContext`, ptal.

linsimiao:
Thank you for the suggestions.

lburgazzoli:
@linsimiao @AdoHe why DnsSrv resolver has been deleted here ?

linsimiao:
@lburgazzoli  I made a stupid mistake here.

lburgazzoli:
@linsimiao no problem :-)

xiang90:
@linsimiao Is this ready for review?

linsimiao:
@xiang90  yup.

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -74,8 +63,10 @@ public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder bui
         this.authClient = newAuthClient(authStub);
         this.maintenanceClient = newMaintenanceClient(mainFStub, mainStub);
         this.clusterClient = newClusterClient(clusterStub);
+
```
AdoHe: no need to add extra line.

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -74,8 +63,10 @@ public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder bui
...
     }
 
+
```
AdoHe: the same.

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -146,24 +137,24 @@ protected EtcdMaintenance getMaintenanceClient(){
...
+        if (config.name != null || config.password != null) {
 
-            checkNotNull(builder.getName(), "username can not be null.");
-            checkNotNull(builder.getPassword(), "password can not be null.");
-            checkArgument(builder.getName().toStringUtf8().trim().length() != 0, "username can not be null.");
-            checkArgument(builder.getPassword().toStringUtf8().trim().length() != 0, "password can not be null.");
+            checkNotNull(config.name, "username can not be null.");
+            checkNotNull(config.password, "password can not be null.");
+            checkArgument(config.name.toStringUtf8().trim().length() != 0, "username can not be null.");
+            checkArgument(config.password.toStringUtf8().trim().length() != 0, "password can not be null.");
```
AdoHe: it's better to put this check in `build` method.

src/main/java/com/coreos/jetcd/EtcdClientBuilder.java
```diff
@@ -3,21 +3,25 @@
...
 /**
  * ClientBuilder knows how to create an EtcdClient instance.
  */
 public class EtcdClientBuilder {
 
     private List<String> endpoints = Lists.newArrayList();
-    private ByteString   name;
-    private ByteString   password;
+    private ByteString name;
+    private ByteString password;
```
AdoHe: I would prefer to use `String` here, and cast it to `ByteString` internal.

src/main/java/com/coreos/jetcd/EtcdClientBuilder.java
```diff
@@ -100,7 +104,7 @@ public EtcdClientBuilder setPassword(ByteString password) {
      * @return this builder
      * @throws NullPointerException if password is null
      */
-    public EtcdClientBuilder setNameResolverFactory(AbstractEtcdNameResolverFactory nameResolverFactory){
+    public EtcdClientBuilder setNameResolverFactory(AbstractEtcdNameResolverFactory nameResolverFactory) {
```
AdoHe: we need to think more about whether we should expose this out.

src/main/java/com/coreos/jetcd/EtcdClientBuilder.java
```diff
@@ -115,6 +119,22 @@ public AbstractEtcdNameResolverFactory getNameResolverFactory() {
...
+        URI[] uris = new URI[endpoints.size()];
+        for (int i = 0; i < endpoints.size(); ++i) {
+            try {
+                String endpoint = endpoints.get(i);
+                if (!endpoint.startsWith("http://")) {
+                    endpoint = "http://" + endpoint;
+                }
+                uris[i] = new URI(endpoint);
+            } catch (URISyntaxException e) {
+                e.printStackTrace();
```
AdoHe: how about `continue` here?

src/main/java/com/coreos/jetcd/EtcdClientBuilder.java
linsimiao: Tks a lot for your reviews, and I will fix the codes.  I think we should use `public` here, cause there are several classes which extend `AbstractEtcdNameResolverFactory`, and sometimes  users may want to use their own implementations, it would be better if we expose this out for them.

fanminshi:
why do we need etcd config when we have clientBuilder already?

fanminshi:
I don't think adding config is useful. ClientBuilder itself acts as a config. @lburgazzoli what do you think?

fanminshi:
I don't think this is needed.

PR #37, opened 2016-08-10
add simple cluster client for etcd

StupidHod:
@xiang90 @AdoHe waiting for your advice, what should I do about cluster?

### DIFF
--- a/src/main/java/com/coreos/jetcd/EtcdClusterImpl.java
+++ b/src/main/java/com/coreos/jetcd/EtcdClusterImpl.java
@@ -1,0 +1,69 @@
+package com.coreos.jetcd;
+
+import com.coreos.jetcd.api.*;
+import com.google.common.util.concurrent.ListenableFuture;
+
+import java.util.List;
+
+/**
+ * Implementation of cluster client
+ */
+public class EtcdClusterImpl implements EtcdCluster {
+
+    private ClusterGrpc.ClusterFutureStub clusterStub;
+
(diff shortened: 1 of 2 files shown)


AdoHe:
Generally LGTM. @StupidHod since test issue has been fixed, we need to setup a cluster with more than one node.

StupidHod:
@AdoHe , OK, I will have a test with cluster, try for one more node.

src/main/java/com/coreos/jetcd/EtcdClusterImpl.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * removes an existing member from the cluster
+     *
+     * @param memberId the id of the member
+     * @return
+     */
+    @Override
+    public ListenableFuture<MemberRemoveResponse> removeMember(long memberId) {
```
xiang90: memberID

src/main/java/com/coreos/jetcd/EtcdClusterImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * update peer addresses of the member
+     *
+     * @param memberId the id of member to update
+     * @param endpoints the new endpoints for the member
+     * @return
+     */
+    @Override
+    public ListenableFuture<MemberUpdateResponse> updateMember(long memberId, List<String> endpoints) {
```
xiang90: memberID

xiang90:
LGTM

PR #38, opened 2016-08-11
add simple maintenance client

StupidHod:
I will add test sample with next commit.

### DIFF
--- a/src/main/java/com/coreos/jetcd/EtcdMaintenance.java
+++ b/src/main/java/com/coreos/jetcd/EtcdMaintenance.java
@@ -1,0 +1,96 @@
+package com.coreos.jetcd;
+
+import com.coreos.jetcd.api.*;
+import com.google.common.util.concurrent.ListenableFuture;
+import io.grpc.stub.StreamObserver;
+
+/**
+ * Interface of maintenance talking to etcd
+ * <p>
+ * An etcd cluster needs periodic maintenance to remain reliable. Depending
+ * on an etcd application's needs, this maintenance can usually be
+ * automated and performed without downtime or significantly degraded
+ * performance.
+ * <p>
+ * All etcd maintenance manages storage resources consumed by the etcd
+ * keyspace. Failure to adequately control the keyspace size is guarded by
+ * storage space quotas; if an etcd member runs low on space, a quota will
+ * trigger cluster-wide alarms which will put the system into a
+ * limited-operation maintenance mode. To avoid running out of space for
+ * writes to the keyspace, the etcd keyspace history must be compacted.
+ * Storage space itself may be reclaimed by defragmenting etcd members.
+ * Finally, periodic snapshot backups of etcd member state makes it possible
+ * to recover any unintended logical data loss or corruption caused by
+ * operational error.
+ */
+public interface EtcdMaintenance {
+
+    /**
+     * get all active keyspace alarm
+     */
+    ListenableFuture<AlarmResponse> listAlarms();
+
+    /**
+     * disarms a given alarm
+     *
+     * @param member the alarm
+     * @return the response result
+     */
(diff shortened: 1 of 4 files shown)


src/main/java/com/coreos/jetcd/EtcdMaintenanceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * disarms a given alarm
+     *
+     * @param member the alarm
+     * @return the response result
+     */
+    @Override
+    public ListenableFuture<AlarmResponse> disAlarm(AlarmMember member) {
```
xiang90: where do we define AlarmMember?

src/main/java/com/coreos/jetcd/EtcdMaintenance.java
```diff
@@ -0,0 +1,70 @@
...
+     * get the status of one member
+     */
+    ListenableFuture<StatusResponse> statusMember();
+
+    /**
+     * This handler will be called when the member make a snapshot
+     *
+     * @param snapshotRspStream
+     */
+    void setSnapshotObserver(StreamObserver<SnapshotResponse> snapshotRspStream);
```
xiang90: I do not quite understand how this method works. Do you have an example?

src/main/java/com/coreos/jetcd/EtcdMaintenanceImpl.java
StupidHod: AlarmMember is returned by listAlarms.

xiang90: We probably should typedef it. We should not directly use the type from protobuf. I know we use them for return values today. But we will get rid of them soon.

src/main/java/com/coreos/jetcd/EtcdMaintenance.java
StupidHod: @xiang90  Ok, I will commit it with a example.

src/main/java/com/coreos/jetcd/EtcdMaintenanceImpl.java
StupidHod: @xiang90 Ok, I will convert it.

src/main/java/com/coreos/jetcd/EtcdMaintenance.java
```diff
@@ -0,0 +1,70 @@
...
+     */
+    ListenableFuture<AlarmResponse> listAlarms();
+
+    /**
+     * disarms a given alarm
+     *
+     * @param member the alarm
+     * @return the response result
+     */
+    ListenableFuture<AlarmResponse> disAlarm(AlarmMember member);
```
hongchaodeng: disalarm

xiang90:
we need to get this rebased

StupidHod:
@xiang90 OK, I have rebased this head to master.

src/main/java/com/coreos/jetcd/EtcdMaintenance.java
xiang90: can we get an example for this? i hope we can get this pr merged soon :)

StupidHod:
I have changed the interface for maintenance and add some tests for maintenance, but the alarm and snapshot is hard for me to trigger so the tests are not good. @xiang90

hongchaodeng:
>  but the alarm and snapshot is hard for me to trigger so the tests are not good
Can we add a TODO for untested methods?

StupidHod:
@hongchaodeng Good idea, I have done it with 'TODO comment' and detail how to use setSnapshotCallback in [testAddSnapshotCallback](https://github.com/coreos/jetcd/pull/38/files#diff-07b4e85b363c509c190262ba07d9c38cR49)

hongchaodeng:
LGTM

PR #43, opened 2016-08-28
scripts: refactoring run_etcd.sh

oza:
1. support to download darwin and mac os based on a result of uname command.
2. skip download binary ball if it exists.
3. upgrade etcd version to v3.0.6.


oza:
I found that run_etcd.sh fails without error if we use OS X. This PR try to fix it.

xiang90:
The test script is mainly for CI. But I think it is fine to add osx support so that dev can run test easier locally. lgtm

Issue #39, opened 2016-08-15
Update grpc version to 1.0.0-pre2?


skyao:
hi, now we are using grpc version 0.15.0, and the new version of grpc is 1.0.0.pre2。

should we update to this version now? One strong reason is that in this version protobuf has been update to release version 3.0.0 and netty is udpate to 3.1.7-final with some bug fixing.

I have to say that in 1.0.0-pre2 (includes pre1) there are some big changes we have to face:

1.  In the generated code from .proto files, all the client interfaces has been moved such as 

    public static interface KV {}
    public static interface KVBlockingClient {}
    public static interface KVFutureClient {}

    PS: in 0.15.0, these interfaces are marked as Deprecated.

2. The stub class is now changed to be "final"

    Now in the client we have only one choice : 
    public static class KVStub extends io.grpc.stub.AbstractStub<KVStub>{}
    And in 1.0.0-pre2, this class is marked in final class.

So if we want to do some unit test to mock the special error case, we have to introduce something like powermock to mock final class.

Should we update it now or later?

xiang90:
already fixed by #44

PR #44, opened 2016-08-28
pom.xml: upgrade grpc and protobuf

oza:
The stable versions of grpc and protobuf are released now.
* grpc: 1.0.0
* protobuf: 3.0.0


oza:
Looks to be related to #39 

My first idea to create this PR is that it's better to update these core libraries at the early stage of jetcd.

xiang90:
lgtm

PR #45, opened 2016-08-30
add etcd watch implementation

StupidHod:
With the wath design, I have finished the watch implementation, hope for your review.
@AdoHe @xiang90


AdoHe:
Good job @StupidHod , will take a look asap.

src/main/java/com/coreos/jetcd/EtcdWatch.java
```diff
@@ -0,0 +1,32 @@
+package com.coreos.jetcd;
+
+import com.coreos.jetcd.options.WatchOption;
+import com.coreos.jetcd.util.ListenableSetFuture;
+import com.coreos.jetcd.watch.Watcher;
+import com.google.protobuf.ByteString;
```
oza: IMHO, it's better to avoid exposing ByteString directly since Apache Felix's iPOJO abstraction can be broken. Let me explain the detail. protobuf compiler generates non-compatible source code(I mean it's source code-level compatibility, not wire-level compatibility). If a user of jetcd tries to use another protobuf compiler(e.g. v2.5.0 compiler) for her applications, it can happen for users to pass non-compatible(v2.5.0) ByteString bytecode. My suggestion is to create a wrapper class of ByteString for preventing users from touching jetcd's ByteString.

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
```diff
@@ -0,0 +1,294 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+/**
+ * etcd watcher Implementation
+ */
+public class EtcdWatchImpl implements EtcdWatch {
+
+    private StreamObserver<WatchRequest> requestStream;
```
oza: For making Double-Checked Locking in ```getRequestStream``` work correctly,  please make ```requestStream``` ```volatile```.

http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
```diff
@@ -0,0 +1,294 @@
...
+     * requestStream automatically
+     * <p>the responseStream will distribute the create, cancel, normal response to
+     * processCreate, processCanceled and processEvents
+     * <p>if error happened, the requestStream will be closed by server side, so we call resume to resume
+     * all ongoing watchers
+     *
+     * @return
+     */
+    private StreamObserver<WatchRequest> getRequestStream() {
+        if (this.requestStream == null) {
```
oza: Based on a previous comment, I think it should be better to avoid double-check locking here. https://github.com/coreos/jetcd/pull/45/files#r76965900

oza:
Hi @StupidHod, thanks for your great job! I'm interested in jetcd, so I'm commenting here. I checked some points about API-level comment and initialization sequence of RequestStream.

src/main/java/com/coreos/jetcd/EtcdWatch.java
AdoHe: @oza thanks for your great suggestion. If we expose java String to users, and convert it to ByteString, wdyt?

oza: One concern to use Java String directly is a semantic gap between Golang's string and Java's string. IIUC, etcd's key can contain illegal String for java String since Golang's string is just a wrapper of byte[]. As a result, it can happen to fail to decode as Java's string. It's not trivial to handle them correctly at jetcd layer. My suggestion is to create primitive APIs which accept/return byte[], and create utility API which convert byte[] in primitive APIs to Java String on the top of them.

AdoHe: @xiang90 could you confirm this? If so we have to wrap this.

xiang90: @oza is correct.

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
StupidHod: @oza Tks for your awesome comments, I can define it as volatile, and this will works well under  JDK 1.5 and later. 
https://www.securecoding.cert.org/confluence/display/java/LCK10-J.+Use+a+correct+form+of+the+double-checked+locking+idiom
(described in method "Compliant Solution (Volatile)")
We use JDK 1.8 in this project, so I think volatile will solve this issue. Thanks for your comments, it really make a lot for me.

xiang90:
@oza Great news! Would love to have you contribute more to jetcd! Thanks a lot for the help so far!

oza:
@xiang90 Happy to hear that :-) Yes, I will contribute more to jetcd!

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
lburgazzoli: As guava is a dependency, what about using Suppliers.memoize ?

AdoHe:
@StupidHod @oza I would support we construct kind of readable/writable byte sequence, we can name it `ByteSequence`, this `ByteSequence` can accpet/return byte[], also can accept string, and if possible we can return string.

lburgazzoli:
ByteSequence should also accept CharSequence

StupidHod:
@AdoHe @lburgazzoli Got your advices.

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
oza: Sounds interesting choise. Can Suppilier.memoize forget previous instance of RequestStream?

lburgazzoli:
No it is a lazy singleton

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
AdoHe: generally, I think `volatile` here is enough, I don't want to make this giantic complicated. Suppliers.memorize is not that common used.

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
```diff
@@ -0,0 +1,294 @@
...
+/**
+ * etcd watcher Implementation
+ */
+public class EtcdWatchImpl implements EtcdWatch {
+
+    private StreamObserver<WatchRequest> requestStream;
+
+    private ConcurrentHashMap<Long, Watcher> watchers = new ConcurrentHashMap<>();
+
+    private WatchGrpc.WatchStub watchStub;
```
AdoHe: `private final` could be better.

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
```diff
@@ -0,0 +1,294 @@
...
+     * If the watch is slow or the required rev is compacted, the watch request
+     * might be canceled from the server-side and the onCreateFailed will be called.
+     *
+     * @param key         the key subscribe
+     * @param watchOption key option
+     * @param callback    call back
+     * @return ListenableFuture watcher
+     */
+    @Override
+    public synchronized ListenableSetFuture<Watcher> watch(ByteString key, WatchOption watchOption, Watcher.WatchCallback callback) {
```
AdoHe: any reason to make this method `synchronized`?

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * Cancel the watch task with the watcher, the onCanceled will be called after successfully canceled.
+     *
+     * @param watcher the watcher to be canceled
+     */
+    @Override
+    public void cancelWatch(Watcher watcher) {
+        Watcher temp = watchers.get(watcher.getWatchID());
+        if (temp != null) {
+            synchronized (temp) {
```
AdoHe: I would doubt synchronize on local variable functionality.

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
```diff
@@ -0,0 +1,294 @@
...
+                if (this.watchers.containsKey(temp.getWatchID())) {
+                    this.watchers.remove(temp.getWatchID());
+                    this.cancelWatchers.put(temp.getWatchID(), temp);
```
AdoHe: is this right to put temp here?

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
```diff
@@ -0,0 +1,294 @@
...
+     */
+    private void resume() {
+        synchronized (this) {
+            this.requestStream = null;
+            Watcher[] resumeWatchers = (Watcher[]) watchers.values().toArray();
+            this.watchers.clear();
+            this.cancelWatchers.clear();
+            resumeWatchers(resumeWatchers);
+        }
+
```
AdoHe: no need extra newline.

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
```diff
@@ -0,0 +1,294 @@
...
+            }
+        }
+    }
+
+    /**
+     * empty the old request stream, watchers and resume the old watchers
+     * empty the cancelWatchers as there is no need to cancel, the old request stream has been dead
+     */
+    private void resume() {
+        synchronized (this) {
```
AdoHe: since the whole code block is synchronized, you could make this method synchronized.

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
```diff
@@ -0,0 +1,294 @@
...
+     * <p>if error happened, the requestStream will be closed by server side, so we call resume to resume
+     * all ongoing watchers
+     *
+     * @return
+     */
+    private StreamObserver<WatchRequest> getRequestStream() {
+        if (this.requestStream == null) {
+            synchronized (this) {
+                if (this.requestStream == null) {
+                    StreamObserver<WatchResponse> watchResponseStreamObserver = new StreamObserver<WatchResponse>() {
```
AdoHe: what if we make this variable instance scope?

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
lburgazzoli: I had a more detailed look at EtcdWatchImpl class and IMHO the DCL can be avoided as getRequestStream is invoked by watch which is synchronized, requestStream is then re-set to null in resume which is actually synchronized too.

src/main/java/com/coreos/jetcd/EtcdWatch.java
```diff
@@ -0,0 +1,32 @@
...
+     * @return ListenableFuture watcher
+     */
+    ListenableSetFuture<Watcher> watch(ByteString key, WatchOption watchOption, Watcher.WatchCallback callback);
+
+    /**
+     * Cancel the watch task with the watcher, the onCanceled will be called after successfully canceled.
+     *
+     * @param watcher the watcher to be canceled
+     */
+    void cancelWatch(Watcher watcher);
```
lburgazzoli: Is this method really needed ? I would expect that as watch returns a future I can call future.cancel(boolean)

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
StupidHod: Ok, got it.

src/main/java/com/coreos/jetcd/EtcdWatch.java
StupidHod: @lburgazzoli The future is to wait the watcher object, the watcher do not implement the future interface, so I think we neeed a function to cancel it.

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
StupidHod: @AdoHe I think it works, as the the lock is to lock the instance but not the reference, we new watcher just once in onProcessCreate, so we always use the same watcher instance, it's not a local variable.

src/main/java/com/coreos/jetcd/EtcdWatch.java
lburgazzoli: Ah correct, so it would be better imho to make Watcher Closeable

src/main/java/com/coreos/jetcd/watch/Watcher.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    public boolean isCanceled() {
+        return canceled;
+    }
+
+    public synchronized void setCanceled(boolean canceled) {
+        this.canceled = canceled;
+    }
+
```
lburgazzoli: I doubt setLastRevision/setCanceled/setResuming should be public

StupidHod: Yes, you are right, it's reasonable to set them as package visible. Is it ok to set the watcher as a internal class of EtcdWatch, as if we set them as protected/private, the EtcdWatch can not see them?
I doubt about this, so I set them as public.

src/main/java/com/coreos/jetcd/EtcdWatch.java
StupidHod: To close it by itself, the watcher need to hold a reference of WatchClient. To close the watcher, the watcher have to call member function of WatchClient. I think it's a little strange.

src/main/java/com/coreos/jetcd/watch/Watcher.java
lburgazzoli: Package private is ok but maybe this can be split in interface/impl ?

src/main/java/com/coreos/jetcd/EtcdWatch.java
lburgazzoli: The problem with this implementation is that you need to carry the EtcdWatch instance that creates the Watcher  with you making it hard to delegate Watch operation. 

Maybe by splitting Watcher and WatcherImpl you can solve the problem, i.e. WatcherImpl could be an EtcdWatch inner class.

src/main/java/com/coreos/jetcd/watch/Watcher.java
StupidHod: Good idea to split in interface/impl, tks for your awesome advice.

StupidHod:
As the discussion, I have do some changes as below.

1. Change some sync detail, such as remove sync on watch function.

2. Split watcher in Watcher(interface in EtcdWatch) and WatcherImpl(class in EtcdWatchImpl).

3. Add cancel function for Watcher, return a CompletableFuture result.

4. Add ByteSequence to replace ByteString

5. Add Etcd Client KeyValue to replace gRPC KeyValue

6. Add Etcd Client Header to replace gRPC Response Header

7. Add EtcdUtil to convert gRPC object to EtcdClient Object.

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
StupidHod: I'm not sure if this object hold any status, so I new one for new requestStream.

src/main/java/com/coreos/jetcd/EtcdClient.java
```diff
@@ -68,12 +70,15 @@ public EtcdClient(ManagedChannelBuilder<?> channelBuilder, EtcdClientBuilder bui
...
             clusterStub = setTokenForStub(clusterStub, token);
+            watchStub = setTokenForStub(watchStub, token);
         }
 
         this.kvClient = newKVClient(kvStub);
         this.authClient = newAuthClient(authStub);
         this.maintenanceClient = newMaintenanceClient(mainFStub, mainStub);
         this.clusterClient = newClusterClient(clusterStub);
+        this.watchClient = newWatchClient(watchStub);
+
```
lburgazzoli: It is really needed to create so many objects bu default ? 
I would instead use Suppliers.memorize or any other way to have lazy object creation

src/main/java/com/coreos/jetcd/EtcdWatch.java
```diff
@@ -0,0 +1,70 @@
...
+         * @return watch option
+         */
+        WatchOption getWatchOption();
+
+        /**
+         * cancel the watcher
+         * @return cancel result
+         */
+        CompletableFuture<Boolean> cancel();
+    }
```
lburgazzoli: I would implement also Closeable

StupidHod: This is to close.

lburgazzoli: Yep but having it as Closeable would make it easy to close it along with other non etcd resources and not having to do something specific, Closeable could be implemented as default method wrapping cancel

StupidHod: Ok, you are right, I will do it.

lburgazzoli: Not a big issue indeed ;)
Just trying to make my future life easy ;)

src/main/java/com/coreos/jetcd/EtcdUtil.java
```diff
@@ -0,0 +1,80 @@
...
+import com.coreos.jetcd.watch.WatchEvent;
+import com.google.protobuf.ByteString;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This util is to convert api class to client class.
+ */
+public class EtcdUtil {
```
oza: How about making EtcdUtil package private? I think this is only used in jetcd internal.

src/main/java/com/coreos/jetcd/data/ByteSequence.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Arrays;
+
+/**
+ * Etcd binary bytes, easy to convert between byte[], String and ByteString.
+ */
+public class ByteSequence {
+
+    private final byte[] bytes;
+    private final int hashVal;
+    private final ByteString byteString;
```
oza: I prefer to remove bytes[] here.

Currently, both of byte[] and ByteString are hold in ByteSequence.  I think there are no need to hold byte[] since all method only use ByteString except ```calcHashCore```.  However, ```calcHashCore``` can use ```source``` and I think it's enough. What do you think?

StupidHod: OK, you are right.

src/main/java/com/coreos/jetcd/EtcdUtil.java
StupidHod: Yes, I have mark it as package private.

AdoHe:
@StupidHod @oza @lburgazzoli I think this is just what we want. All you guys are amazing 👍 For this PR I would hope we can focus on the thread safety, I would like to implement something quite clean but enough safe.

StupidHod:
Yes, for some points, we can implement them later.

oza:
OK, checking thread safety.

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
```diff
@@ -0,0 +1,415 @@
...
+                    this.watchers.remove(temp.getWatchID());
+                    completableFuture = new CompletableFuture<>();
+                    this.pendingCancelFutures.put(id, completableFuture);
+                }
+            }
+        }
+
+        WatchCancelRequest cancelRequest = WatchCancelRequest.newBuilder().setWatchId(id).build();
+        WatchRequest request = WatchRequest.newBuilder().setCancelRequest(cancelRequest).build();
+        this.requestStream.onNext(request);
```
oza: ```requestStream``` can be null without lock. ```getRequestStream()``` should be called instead.

StupidHod: Thanks very much.

sundbp:
Curious if this will get merged soon? watch implementation and a release of jetcd being available would be great!

AdoHe:
@StupidHod @oza @lburgazzoli let's check what still remaining, we should get this merge asap.

StupidHod:
@AdoHe I think you are right, I have fixed the issues proposed, hope for your advices.

abhin4v:
@AdoHe Will this PR get merged soon? Last comment on this was almost a month ago.

AdoHe:
@abhin4v I will recheck this, and believe we can merge this before this weekend.

blackgwe:
@xiang90 Any show stopper for this PR?

AdoHe:
@blackgwe I will recheck this with @StupidHod, will get this merged asap.

StupidHod:
OK, got it. @AdoHe

forif:
We now urge for the java watcher service,  hope this PR get merged soon.

xiang90:
@AdoHe Can you please take another look?

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
```diff
@@ -0,0 +1,417 @@
+package com.coreos.jetcd;
+
+import com.coreos.jetcd.api.*;
+import com.coreos.jetcd.data.ByteSequence;
+import com.coreos.jetcd.options.WatchOption;
+import com.coreos.jetcd.watch.WatchCreateException;
+import com.google.protobuf.ByteString;
+import io.grpc.ManagedChannel;
+import io.grpc.stub.StreamObserver;
+import javafx.util.Pair;
```
krestenkrab: You should not include `javafx` as a dependency< nothing else in this project does. It also means this only works with Oracle JDKs (i.e. it won't work with OpenJDK).  If need be, add a package-local class at the end of the file like this:

````
class Pair<K,V> {
    final K k;
    final V v;
    Pair(K k, V v) { this.k = k; this.v = v; }
    K getKey() { return k; }
    V getValue() { return v; }

    @Override
    public int hashCode() {
        return k.hashCode() + v.hashCode();
    }

    @Override
    public boolean equals(Object obj) {
        if (obj==null || !(obj instanceof Pair)) return false;
        Pair<K,V> other = (Pair<K,V>)obj;
        return k.equals(other.k) && v.equals(other.v);
    }
}
````

File: src/main/java/com/coreos/jetcd/EtcdWatchImpl.java [hunk 19]
```diff
@@ -0,0 +1,417 @@
...
+        return completableFuture;
+    }
+
+    /**
+     * empty the old request stream, watchers and resume the old watchers
+     * empty the pendingCancelFutures as there is no need to cancel, the old request stream has been dead
+     */
+    private synchronized void resume() {
+        this.requestStream = null;
+        WatcherImpl[] resumeWatchers = (WatcherImpl[]) watchers.values().toArray();
```
krestenkrab: This should be

````
        WatcherImpl[] resumeWatchers = (WatcherImpl[]) watchers.values().toArray(new WatcherImpl[watchers.size()]);
````

or you will get a class cast exception

src/main/java/com/coreos/jetcd/EtcdWatchImpl.java
krestenkrab: This comment applies for the https://github.com/enncloud/jetcd changes too

StupidHod:
@krestenkrab Thank you for your reviewing. I will fix them  soon.

Issue #46, opened 2016-09-01
EtcdLease should not use ListenableFuture directly


oza:
Currently, [EtcdLease](https://github.com/coreos/jetcd/blob/8658eb43e975e69d0030d71147323b5ef37ecb41/src/main/java/com/coreos/jetcd/EtcdLease.java) exposes com.google.common.util.concurrent.ListenableFuture as a part of API. If ListenableFuture is used directly, it will conflict with user-side guava library. Instead, [ListenableSetFuture](https://github.com/StupidHod/jetcd/blob/8d4d09c78aa6029bfabbb3a6cd9c6120c38ba5dd/src/main/java/com/coreos/jetcd/util/ListenableSetFuture.java) in #45 or some wrapper class in com.coreos.jetcd should be used to avoid the problem.

lburgazzoli:
This is related to #16.
IMHO best would be to shade guava and use CompletableFutures as suggested by @jurmous

oza:
@lburgazzoli thanks for the notification. I will also check #16.

I have a patch to shade Guava into jetcd locally, so I can send a PR soon. However, it can increase the size of binary. Can we accept it?

lburgazzoli:
I do not know, but having had quite a few issues with guava dependencies I would really like not to do it again ;)

oza:
Thanks for your comment. I agree with you. Created #47. I appreciate if you take look.

fanminshi:
closing via https://github.com/coreos/jetcd/pull/109

PR #47, opened 2016-09-01
Shading Guava.

oza:
This PR tries to solve some problems described in #46 and #16. grpc and protobuf can also be shaded here.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,18 +20,26 @@
         <maven-source-plugin.version>3.0.1</maven-source-plugin.version>
         <maven-compiler-plugin.version>3.5.1</maven-compiler-plugin.version>
         <maven-resources-plugin.version>3.0.1</maven-resources-plugin.version>
+        <maven-remote-resources-plugin.version>1.5</maven-remote-resources-plugin.version>
+        <maven-shade-plugin.version>2.4.3</maven-shade-plugin.version>
         <maven-javadoc-plugin.version>2.10.4</maven-javadoc-plugin.version>
         <maven-bundle-plugin.version>3.2.0</maven-bundle-plugin.version>
         <depends-maven-plugin.version>1.3.1</depends-maven-plugin.version>
         <os-maven-plugin.version>1.5.0.Final</os-maven-plugin.version>
         <protobuf-maven-plugin.version>0.5.0</protobuf-maven-plugin.version>
 
+        <!-- -->
+        <etcd.shade.packageName>com.coreos.jetcd.repackaged</etcd.shade.packageName>
+
         <!-- dependencies -->
         <grpc.version>1.0.0</grpc.version>
         <slf4j.version>1.7.21</slf4j.version>
 
+
         <!-- should be in sync with protobuf grpc dependencies -->
         <protobuf.version>3.0.0</protobuf.version>
+        <!-- should be in sync with grpc dependencies -->
+        <guava.version>19.0</guava.version>
 
         <!-- test dependencies -->
         <testng.version>6.9.10</testng.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -65,6 +73,12 @@
             <artifactId>slf4j-api</artifactId>
             <version>${slf4j.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>${guava.version}</version>
```
sijie: since gRpc depends on guava, I think you don't have to list the exact version here, no?

sijie:
@oza I am actually not sure shading a dependency for a public interface is right solution. for example, if my project uses guava ListenableFuture and etcd exposes a re-packaged ListenableFuture. It will become a bit weird because they are all *guava* ListenableFuture. However, I can't simply chain the callbacks, I have to write another adapter to covert ectd re-packaged ListenableFuture to guava ListenableFuture.

I would suggest not shade in etcd. If the application has dependency conflicts with the dependencies that jetcd has, it won't be just simply guava version conflicts. There will be other dependency conflicts. so this should really leave application to resolve the conflicts rather than the library itself.

oza:
@sijie @xiang90 make sense. IMO, it's better to avoid shading since we have iPOJO abstraction and we can easily avoid conflict.Shading can cause a problem to confuse users because of same name of classes, to make fat binary.

@lburgazzoli what do you think? Fortunately, jetcd has not yet decided the interfaces. Hence, we can change the interfaces instead of shading.

sijie:
I would say lets use the current ListenableFuture to complete the first version of jetcd. When we approach the first release version for jetcd, lets decide what is the best *Future*/*Callback* interface. does that make sense to everyone here?

lburgazzoli:
Shading guava would make sense if guava classes are not exposed to public api. 

I would say that such decision of how interfaces should look like is something to address as early as possible to avoid breaking changes after release.

oza:
> I would say lets use the current ListenableFuture to complete the first version of jetcd. When we approach the first release version for jetcd, lets decide what is the best Future/Callback interface. does that make sense to everyone here?

As the same reason that we shouldn't expose ByteString of protobuf to end-user [here](https://github.com/coreos/jetcd/pull/45#discussion_r76964462), I think it's not good option. It's OK to use ListenableFuture internally. However, if we mix bytecode of different version Guava, it can cause method missing error. I also breaks iPOJO class path isolation by Apache Felix. To avoid them, we can have a 2 choices:

1. Changing interface.
2. Shading Guava without changing interface.

Please correct me if I'm wrong.

sijie:
I am okay with changing to java Future and byte[] or ByteBuffer, which are those standard libs.

lburgazzoli:
Wouldn't CompletableFuture be better ?

oza:
@lburgazzoli CompletableFuture seems to be better to me.

AdoHe:
Agree. I think it's time for use to make a decision, I would support use CompletableFuture and custom ByteSequence.

sijie:
CompletableFuture works for me. why not java nio ByteBuffer?

oza:
@sijie Good catch. I would like to suggest that ByteSequence accept both ByteBuffer and byte[]. It's because that both of ByteBuffer and byte[] are used for buffer management(e.g. protobuf's accepts both of them).

I think ByteBuffer is used for safe buffer management by switching read/write mode, and byte[] is used for its simplicity.

fanminshi:
I replace usage of ByteString from all exposed API to ByteSequence in https://github.com/coreos/jetcd/pull/47. I think we don't shading guava library anymore since jetcd APIs doesn't depends on guava.

PR #48, opened 2016-09-05
pom.xml: Enabling checkstyle at compile time

oza:
This PR enables checkstyle at compile time with maven-checkstyle-plugin. In this PR, it uses google_checks.xml as coding style, but it can be changed based on discussion here.


xiang90:
lgtm

PR #51, opened 2016-09-15
fix get methods access issue

AdoHe:
those getter methods should be `public`.


lburgazzoli:
@AdoHe 

Why all the new*Client have been made private ? 

In addition I would make all the get*Client lazy i.e. wi th Suppliers.memorize so the client do not allocate unused objects.

bkrahmer:
I was about to add an issue for the same thing before a colleague spotted this pull request.  Can you please merge it?  thanks!

AdoHe:
@bkrahmer merged, ptal.

lburgazzoli:
@AdoHe so the new* are all private and clients are created inside the contructor ?

PR #65, opened 2016-12-21
Bump to Etcd 3.0.15

oscerd:


oscerd:
any interest in this? I can close if not.

fanminshi:
the current etcd version is v3.1.3. Probably should bump to that version. @xiang90 what do you think?

xiang90:
let's bump to the latest

PR #68, opened 2017-02-13
format: format code to remove indentation warning

fanminshi:


fanminshi:
only style change. merging

PR #69, opened 2017-02-13
format: reorder imports

fanminshi:


fanminshi:
this only solve partial reordering warning but not all.

xiang90:
lgtm

PR #72, opened 2017-02-23
Front page instructions on how to run the tests

akauppi:
Crucial information on how newcomers can check that the code base is healthy on their system.

I took a big bunch of console output in. Personally I find it good to see what I'm expected to see when tests run, especially when a project is in a young stage like this one. This block can of course be trimmed down.

- Wouldn't it be nicer to run the etcd cluster in Docker, instead of natively via shell script?
- That might also make things Windows compatible (now macOS and Linux are supported by the shell script)

I personally find Docker more ... trustworthy than a shell script that loads quite a lot off the Internet.

- Is it necessary to run the first cluster node off port `2379`? If it were, say `12379`, chances of conflicting with an existing installation would be virtually zero.

Can make Issues on the above entries, but writing these instructions brought them up, so wanted to ask first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@ target/
 logs/
 *.releaseBackup
 *.versionsBackup
+
+# Created by './scripts/run_etcd.sh'
+/external/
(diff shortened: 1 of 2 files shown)


xiang90:
LGTM.

We do not have this issue in our go client since we can start an etcd cluster in a go program. We can explore docker approach for java client testing.

heyitsanthony:
lgtm

PR #74, opened 2017-03-18
style: fix style warnings

fanminshi:
fixed all style warnings and allows build to fail on future style warning.


heyitsanthony:
lgtm

PR #75, opened 2017-03-21
add support to grpc 1.1.2

overstart:


fanminshi:
etcd still uses grpc-go 1.04,https://github.com/coreos/etcd/blob/master/glide.yaml#L100. I don't want to upgrade to a grpc version that's higher than that of etcd yet.

xiang90:
@fanminshi I am not sure how java grpc version their client. Can you check if java 1.1 is suggested to work with grpc server 1.1 only, not 1.0?

fanminshi:
I didn't find anything from google search. So I created a issue on grpc-go, https://github.com/grpc/grpc-java/issues/2844

fanminshi:
@xiang90 grpc people says that 

"The versioning of gRPC is called semantic versioning. All major versions, i.e. the leading number, that are the same will work together."

Also the latest is grpc-java 1.2.0, https://github.com/grpc/grpc-java/releases/tag/v1.2.0 but uses protobuf 3.2.0. We might want to upgrade to the latest.

xiang90:
sgtm.

pom.xml
```diff
@@ -27,7 +27,7 @@
         <protobuf-maven-plugin.version>0.5.0</protobuf-maven-plugin.version>
 
         <!-- dependencies -->
-        <grpc.version>1.0.0</grpc.version>
+        <grpc.version>1.1.2</grpc.version>
```
fanminshi: Could you use 1.2.0 for `grpc.version`and 3.2 for `protobuf.version`?

fanminshi:
can you also change the git commit message to
`pom: add support to grpc 1.2.0 and protobuf 3.2.0`

fanminshi:
close via #77

PR #79, opened 2017-04-03
Add new features: getWithPrefixe and deleteWithPrefix

ikonglong:
1. Get all the keys that match the given prefix
2. Delete all the keys that match the given prefix

### DIFF
--- a/src/main/java/com/coreos/jetcd/util/KeyPrefixUtil.java
+++ b/src/main/java/com/coreos/jetcd/util/KeyPrefixUtil.java
@@ -1,0 +1,39 @@
+package com.coreos.jetcd.util;
+
+import com.google.protobuf.ByteString;
+
+import java.util.Arrays;
+
+public abstract class KeyPrefixUtil {
+
+  private static final byte[] NO_PREFIX_END = {0};
+
+  /**
+   * Gets the range end of the given prefix.
+   *
+   * <p> The range end is the key plus one (e.g., "aa"+1 == "ab", "a\xff"+1 == "b").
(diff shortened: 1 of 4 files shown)


src/main/java/com/coreos/jetcd/util/KeyPrefixUtil.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Gets the range end of the given prefix.
+   *
+   * <p>The range end is the key plus one (e.g., "aa"+1 == "ab", "a\xff"+1 == "b").
+   *
+   * @param prefix the given prefix
+   * @return the range end of the given prefix
+   */
+  public static final ByteString prefixEndOf(ByteString prefix) {
```
fanminshi: I think the following code is enough.

```Java
static final ByteString prefixEndOf(ByteString prefix) {
    byte[] endKey = prefix.toByteArray().clone();
    for (int i = endKey.length - 1; i >= 0; i--) {
      if (endKey[i] < 0xff) {
        endKey[i] = (byte) (endKey[i] + 1);
        return ByteString.copyFrom(Arrays.copyOf(endKey, i + 1));
      }
    }

    return ByteString.copyFrom(NO_PREFIX_END);
  }
```

fanminshi:
can you change the commit msgs to be:

`options: add prefixEndOf and tests`
`style: adjust the placement of the static import statements`

fanminshi:
close this for https://github.com/coreos/jetcd/pull/99

PR #82, opened 2017-04-17
Provide a more friendly transaction builder

emiapwil:
Regarding this issue: https://github.com/coreos/jetcd/issues/60

Signed-off-by: Emiapwil (Gao Kai) <emiapwil@gmail.com>


heyitsanthony:
should this include a test so the functionality isn't unintentionally clobbered in the future?

xiang90:
this change allows things like:

if(x).if(x).then(x).if(x).then(x), which does not make any sense.

for each builder, we should allow exact one if, else, then. we should throw exception otherwise.

emiapwil:
Point taken. I will do that in a few days.

fanminshi:
@emiapwil could address @xiang90's concern and add few tests for this?

fanminshi:
close this via https://github.com/coreos/jetcd/pull/101

PR #94, opened 2017-05-12
Update EtcdUtil.java byteSequceFromByteString

raoofm:


heyitsanthony:
change the first line of the commit to `rename byteSequceFromByteString to byteSequenceFromByteString` so it describes what the commit is doing and this should be OK to merge. Thanks!

raoofm:
@heyitsanthony thanks, done

PR #98, opened 2017-05-22
etcdwatch: clean up watch code

### LINKED ISSUE #86: Make watch API stable
- [ ] Revisit watch API to make it consistent with etcd/clientv3 Watch API
- [ ] Add test

fanminshi:
Fixes #86


fanminshi:
Working in progress.

riclava:
why travis ci build failed?

fanminshi:
@riclava my unit test for watch is failing. need to take a look.

src/main/java/com/coreos/jetcd/EtcdWatch.java
```diff
@@ -1,74 +1,81 @@
...
-    void onWatch(EtcdHeader header, List<WatchEvent> events);
-
-    /**
-     * onResuming will be called when the watcher is on resuming.
+     * @throws InterruptedException if interrupted while waiting.
+     * @throws ConnectException if watch client experiences unrecoverable connection error.
+     * @throws TimeoutException if watcher timeout on poll.
      */
-    void onResuming();
+    WatchResponse poll(long timeout, TimeUnit unit)
```
xiang90: how about removing take? only keep poll?

also poll only takes one parameter as duration in nanoseconds?

ncubrian: Could you consider adding withPrefix option to WatchOption?

xiang90: that is when you create the watch, not when you do the polling.

ncubrian: @xiang90 I know. I'm just wondering if you have a plan on adding watch with prefix.

fanminshi: @ncubrian I will add watch with prefix option. Also currently, watch do support prefix behavior. you can use `withRange(ByteSequence endKey)` where endkey is one bit larger than the given key, then it gets all keys with the prefix.

src/main/java/com/coreos/jetcd/exception/CompactedException.java
```diff
@@ -0,0 +1,20 @@
+package com.coreos.jetcd.exception;
```
xiang90: can we add etcd exception in another pr?

fanminshi: yeah I can.

xiang90:
actually the first few prs can be separated from this one. can you also do that. keeping each pr small is the goal.

fanminshi:
I have clean up the watch. PTAL.

fanminshi:
cleaned up this pr. ptal.

src/main/java/com/coreos/jetcd/data/Header.java
```diff
@@ -9,15 +9,12 @@
   private final long memberId;
   private final long revision;
   private final long raftTerm;
-  private final long compactRevision;
```
xiang90: can you also move this commit to a separate pr?

src/main/java/com/coreos/jetcd/internal/impl/Util.java
```diff
@@ -172,7 +173,12 @@ static DeleteResponse toDeleteResponse(DeleteRangeResponse rangeResponse) {
    * convert API CompactionResponse to client CompactResponse.
    */
   static CompactResponse toCompactResponse(CompactionResponse response) {
-    return new CompactResponse(toHeader(response.getHeader(), 0));
+    return new CompactResponse(toHeader(response.getHeader()));
+  }
+
+  static WatchResponse toWatchResponse(com.coreos.jetcd.api.WatchResponse response) {
```
xiang90: also this commit should be in a separate pr.

src/main/java/com/coreos/jetcd/watch/WatchCreateException.java
```diff
@@ -1,16 +0,0 @@
-package com.coreos.jetcd.watch;
```
xiang90: this should be in a separate pr.

src/main/java/com/coreos/jetcd/options/WatchOption.java
```diff
@@ -98,11 +97,6 @@ public Builder withNoPut(boolean noPut) {
       return this;
     }
 
-    public Builder withResuming(boolean resuming) {
```
xiang90: this can be done in a separate pr

src/main/java/com/coreos/jetcd/watch/WatchCreateException.java
fanminshi: WatchCreateException is used in the current(not this pr) watch impl. It will break the ci if i just delete this exception.

src/main/java/com/coreos/jetcd/options/WatchOption.java
fanminshi: yep, can do.

src/main/java/com/coreos/jetcd/watch/WatchCreateException.java
xiang90: ok. then keep it here.

src/main/java/com/coreos/jetcd/data/Header.java
fanminshi: header is tied to watch impl. this change will break ci. it must be changed along with watch impl.

fanminshi: Maybe I don't change Header to include in this pr since compactRevision doesn't break anything. Once the watch pr is merged then I will have another pr to remove compactRevision.

src/main/java/com/coreos/jetcd/internal/impl/Util.java
fanminshi: yep will do.

src/main/java/com/coreos/jetcd/options/WatchOption.java
fanminshi: take a look. it is tied with watch impl. I can't remove it alone.

src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
```diff
@@ -1,44 +1,69 @@
...
+  private static final Logger logger = Logger.getLogger(WatchImpl.class.getName());
 
   private final WatchGrpc.WatchStub watchStub;
 
-  private ConcurrentLinkedQueue<Pair<WatcherImpl, CompletableFuture<Watcher>>>
-      pendingCreateWatchers = new ConcurrentLinkedQueue<>();
-  private Map<Long, CompletableFuture<Boolean>> pendingCancelFutures = new ConcurrentHashMap<>();
+  private volatile StreamObserver<WatchRequest> grpcWatchStreamObserver;
+
+  private final ConcurrentHashMap<Long, WatcherImpl> watchers = new ConcurrentHashMap<>();
```
xiang90: what is the key of the hash map? document it?

fanminshi: Lease-> WatcherImpl. and will document.

src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
```diff
@@ -48,234 +73,275 @@
...
+    Preconditions.checkState(!isClosed(), "Watch client has been closed");
+
+    this.setClosed();
+    this.notifyWatchers(EtcdExceptionFactory.newEtcdException("Watch client has been closed"));
+    this.closeGrpcWatchStreamObserver();
+    this.executor.shutdownNow();
+    this.scheduledExecutorService.shutdownNow();
+  }
+
+  private void notifyWatchers(EtcdException e) {
```
xiang90: closeAllWatchersWithError?

fanminshi: the log message is not accurate. the client should notify the watchers with a exception but not closing it. user is responsible for closing watcher.

src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
```diff
@@ -48,234 +73,275 @@
...
+    return status.getCode() != Status.Code.UNAVAILABLE && status.getCode() != Status.Code.INTERNAL;
+  }
+
+  private void processCreate(WatchResponse response) {
+    WatcherImpl watcher = this.pendingWatchers.poll();
+
+    this.sendNextWatchCreateRequest();
+
+    if (watcher == null) {
+      // shouldn't happen
```
xiang90: throw exception? not sure what is the best practice for java to panic stuff.

src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
fanminshi: so notifyWatchers only notifies watchers but close them.

xiang90:
The interface is much cleaner. The impl seems ok. We can get this merged. I want to add some integration tests after this PR.

src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
xiang90: why not close the watcher for the users?

src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
fanminshi: In the case of duplicate create response. this situation can happen but it shouldn't. I think log this case will be fine since correctness is not compromised by duplicate create messages.

src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
```diff
@@ -48,234 +73,275 @@
...
+    }
+    this.grpcWatchStreamObserver.onCompleted();
+    this.grpcWatchStreamObserver = null;
+  }
+
+  private boolean isNoLeaderError(Status status) {
+    if (status.getDescription() == null) {
+      return false;
+    }
+    return status.getDescription().contains("no leader");
```
fanminshi: I can build a map of etcd server error to jetcd error. then use that map to determine "no leader" error. I can do that in a separate pr.

fanminshi:
comments addressed. PTAL.

xiang90:
lgtm in general.

PR #99, opened 2017-05-24
options: add withPrefix()

fanminshi:
merge #79


src/test/java/com/coreos/jetcd/EtcdKVTest.java
```diff
@@ -146,6 +168,27 @@ public void testDelete() throws Exception {
     }
   }
 
+  @Test(dependsOnMethods = "testGetWithPrefix")
```
xiang90: not sure if this is a java practice. but i am not a fan of introducing dependencies into tests.

xiang90:
LGTM

src/test/java/com/coreos/jetcd/EtcdKVTest.java
fanminshi: it is not a good practice. I'll change it.

PR #100, opened 2017-05-26
style: remove prefix etcd* from class name

fanminshi:


xiang90:
LGTM if CI passes.

fanminshi:
@xiang90 should I have client suffix like WatchClient, KVClient, LeaseClient..? specifically, there is already a class called Lease that just hold lease information. so Lease (client) conflicts with Lease.

xiang90:
probably rename https://github.com/coreos/jetcd/blob/master/src/main/java/com/coreos/jetcd/lease/Lease.java this thing.

fanminshi:
yeah, it seems to me that's same as `keepAlive` in go client. so i'll change it to that.

PR #101, opened 2017-05-30
Txn: allows chaining of IF, Then, and Else in Txn

### LINKED ISSUE #60: Txn.Builder has unfriendly API
I think Txn.Builder would be more user friendly if it allowed appending If/Then/Else operations rather than just replacing them. Right now you have to prepare them in some lists, then add them to builder.

fanminshi:
Fixes #60

### DIFF
--- a/src/test/java/com/coreos/jetcd/op/TxnTest.java
+++ b/src/test/java/com/coreos/jetcd/op/TxnTest.java
@@ -1,0 +1,43 @@
+package com.coreos.jetcd.op;
+
+import com.coreos.jetcd.options.PutOption;
+import com.google.protobuf.ByteString;
+import org.testng.annotations.Test;
+
+public class TxnTest {
+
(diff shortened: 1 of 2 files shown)


src/test/java/com/coreos/jetcd/op/TxnTest.java
```diff
@@ -0,0 +1,46 @@
...
+  @Test(expectedExceptions = IllegalArgumentException.class, expectedExceptionsMessageRegExp = "cannot call If after Else!")
+  public void testIfAfterElse(){
+    Txn.newBuilder().Else(OP).If(CMP).build();
+  }
+
+  @Test(expectedExceptions = IllegalArgumentException.class, expectedExceptionsMessageRegExp = "cannot call Then after Else!")
+  public void testThenAfterElse(){
+    Txn.newBuilder().Else(OP).Then(OP).build();
+  }
+}
```
xiang90: new line

xiang90:
lgtm. thanks!

PR #102, opened 2017-05-31
maintenance: change snapshot api.

fanminshi:


fanminshi:
work in progress. Will add test if the api seems reasonable.

heyitsanthony:
Complete RPC responses probably don't need to be relayed back to the user since it's mostly metadata about transmitting a file.

An `InputStream`/`OutputStream` with the snapshot payload as the byte stream may be easier to use (e.g., clientv3 returns an io.Reader that only has the payload data) with other java code.

src/main/java/com/coreos/jetcd/Maintenance.java
```diff
@@ -64,36 +66,47 @@
...
+     * end of stream is detected, or an exception is thrown.
+     *
+     * <p>when there are no snapshot data to be read,
+     * a SnapshotReaderResponse with <code>remaining</code> -1 is returned.
+     *
+     * @return @see com.coreos.jetcd.maintenance.SnapshotReaderResponse
+     * @throws IOException if connection issue, SnapshotReader closed, any I/O issues.
      */
-    void onError(Throwable throwable);
+    SnapshotReaderResponse read() throws IOException;
```
xiang90: as @heyitsanthony mentioned, we do not really need to return metadata.

probably we only need to implement the abstract func in https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html.

heyitsanthony: doesn't need to implement `InputStream` directly (looked too complicated, but maybe there's a shortcut?)

it could be something like:
```java
class Snapshot {
    ...
    public void write(OutputStream os) throws IOException;
    ...
};
```
can get an `InputStream` out of `write` by using `PipedInputStream`

xiang90: @heyitsanthony i gave it a quick look. it seems that the only we need to implement is read, which return an int representation of byte? maybe i am wrong. something like what you suggest should also work.

fanminshi: @heyitsanthony seems good. `write(OutputStream os)` should block until it is finished successfully or throw an IOException if there is an error. I think that behavior is simple to understand.

xiang90: sounds good to me.

fanminshi:
I changed api to `void write(OutputStream os) throws IOException`. 
I manually test with following and it worked.
1. start one node etcd
2. `put test snapshot`
3. use `snapshot.write()` to write backend snapshot to `snapshot.db`.
4. use `etcdctl snapshot restore snapshot.db`.
5. restart etcd with `snapshot.db`.
6. `get test` returns successfully.

I'll add more unit test for mocking failure scenario if the api seems right.

xiang90:
lgtm

xiang90:
this fixes https://github.com/coreos/jetcd/issues/76

xiang90:
@fanminshi shall we get this merged?

fanminshi:
@xiang90 I am writting some unit tests. I''ll get them done today and the merge.

fanminshi:
@xiang90 done writing unit tests. Can merge after green.

PR #104, opened 2017-06-01
integration: run AuthClientTest the last with priority = 1

fanminshi:
when AutClient sets auth enable and other tests are running,
this causes clients from other tests to throw auth errors.

let AuthClientTest to run last with priority = 1 aviods above issue.

### DIFF
--- a/src/test/java/com/coreos/jetcd/AuthClientTest.java
+++ b/src/test/java/com/coreos/jetcd/AuthClientTest.java
@@ -51,7 +51,7 @@ public class AuthClientTest {
   /**
    * create role with un-auth etcd client
    */
-  @Test(groups = "role")
+  @Test(groups = "role", priority = 1)
   public void testRoleAdd() throws ExecutionException, InterruptedException {
     this.authClient.roleAdd(roleName).get();
   }
@@ -59,7 +59,7 @@ public class AuthClientTest {
   /**
    * grant permission to role
    */
-  @Test(dependsOnMethods = "testRoleAdd", groups = "role")
+  @Test(dependsOnMethods = "testRoleAdd", groups = "role", priority = 1)
   public void testRoleGrantPermission() throws ExecutionException, InterruptedException {
     this.authClient
         .roleGrantPermission(roleName, keyRangeBegin, keyRangeEnd, Permission.Type.READWRITE).get();
@@ -68,7 +68,7 @@ public class AuthClientTest {
   /**
    * add user with password and username
    */
-  @Test(groups = "user")
+  @Test(groups = "user", priority = 1)
   public void testUserAdd() throws ExecutionException, InterruptedException {
     this.authClient.userAdd(userName, password).get();
   }
@@ -76,7 +76,7 @@ public class AuthClientTest {
   /**
    * grant user role
    */
-  @Test(dependsOnMethods = {"testUserAdd", "testRoleGrantPermission"}, groups = "user")
+  @Test(dependsOnMethods = {"testUserAdd", "testRoleGrantPermission"}, groups = "user", priority = 1)
   public void testUserGrantRole() throws ExecutionException, InterruptedException {
     this.authClient.userGrantRole(userName, roleName).get();
   }
@@ -84,7 +84,7 @@ public class AuthClientTest {
   /**
    * enable etcd auth
    */
-  @Test(dependsOnGroups = "user", groups = "authEnable")
+  @Test(dependsOnGroups = "user", groups = "authEnable", priority = 1)
   public void testEnableAuth() throws ExecutionException, InterruptedException {
     this.authClient.authEnable().get();
   }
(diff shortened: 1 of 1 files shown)


fanminshi:
this fixes ci errors on https://github.com/coreos/jetcd/pull/78

heyitsanthony:
Would it be possible to use `@Before` and `@After` annotations to enable/disable auth for these tests instead of using priorities? This seems like a good case for the usual setup/teardown pattern.

How is junit ensuring auth is being enabled for these tests? I see testEnableAuth but don't understand what's making sure it's called before the other auth tests.

fanminshi:
@heyitsanthony the test is running with `TestNG` not `Junit`, and it is using `dependsOnMethods` or `dependsOnGroups` for test ordering.

`testEnableAuth` belongs to the group `authEnable`. It will run if test cases in `user` group has finished. 
```java
  @Test(dependsOnGroups = "user", groups = "authEnable")
  public void testEnableAuth() throws ExecutionException, InterruptedException {
    this.authClient.authEnable().get();
  }
```

`testKVWithAuth()` belongs to group `testAuth` and will only run if tests belongs to group `authEnable` has finished; `testEnableAuth` must have already run.
```java
 @Test(groups = "testAuth", dependsOnGroups = "authEnable")
  public void testKVWithAuth() throws ExecutionException, InterruptedException {
    Throwable err = null;
    try {
      this.secureClient.getKVClient().put(testKey, testName).get();
      RangeResponse rangeResponse = this.secureClient.getKVClient().get(testKey).get();
      this.test.assertTrue(
          rangeResponse.getCount() != 0 && rangeResponse.getKvs(0).getValue().equals(testName));
    } catch (StatusRuntimeException sre) {
      err = sre;
    }
    this.test.assertNull(err, "KV put range test with auth");
  }
```
`Would it be possible to use @Before and @After annotations to enable/disable auth for these tests instead of using priorities? This seems like a good case for the usual setup/teardown pattern.`

The tests in `AuthClientTest` are constructed in a way that every test depends on some previous tests.

The `testDisableAuth` disable `auth` and `delUser` test immediately checks to see if an regular client(without user and password) `this.client.getAuthClient().userDelete(userName).get()` can delete a user.

I think having ` @Before and @After annotations to enable/disable auth` breaks that flow. 

```java
@Test(dependsOnGroups = "testAuth", groups = "disableAuth")
  public void testDisableAuth() {
    Throwable err = null;
    try {
      this.secureClient.getAuthClient().authDisable().get();
    } catch (Exception e) {
      err = e;
    }
    this.test.assertNull(err, "auth disable");
  }

  /**
   * delete user
   */
  @Test(dependsOnGroups = "disableAuth", groups = "clearEnv")
  public void delUser() {
    Throwable err = null;
    try {
      this.client.getAuthClient().userDelete(userName).get();
    } catch (Exception e) {
      err = e;
    }
    this.test.assertNull(err, "user delete");
  }
```

I don't think we can assume that maven run tests sequentially. so it is possible that a test that is trying to do a `Put` in another test class at the moment that `testEnableAuth` in AuthClientTest was running. If `Put` happens after `testEnableAuth`, then `Put` will fail since the etcd server is auth enabled. By setting priority = 1 to all test cases in `AuthClientTest` ensures all other test cases(with default priority = 0) in other test suite must run before `AuthClientTest`; In this case, the scenario running `Put` at same time as `testEnableAuth` can't happen.

fanminshi:
@heyitsanthony also all the integration tests are running against a 3 nodes cluster via https://github.com/coreos/jetcd/blob/master/scripts/run_etcd.sh. That's why there issues with integration tests that ordering can affect correctness of the tests. I think we need to rework the integration tests like that one we have for etcd clientv3. Probably will do that in a future release.

heyitsanthony:
> The tests in AuthClientTest are constructed in a way that every test depends on some previous tests.

OK, but typically unit tests are meant to be small/isolated/self-contained, instead of depending on a long chain of tests. I don't think it's especially worth preserving these dependencies. To wit:

```java
  @Test(dependsOnGroups = "disableAuth", groups = "clearEnv")
  public void delUser() {
    Throwable err = null;
    try {
      this.client.getAuthClient().userDelete(userName).get();
    } catch (Exception e) {
      err = e;
    }
    this.test.assertNull(err, "user delete");
  }
```

Why does this depend on `disableAuth` if it's using the auth client? What about a test for deleting users with auth enabled? etc etc...

The priority stuff is OK as a hack to get stuff working (is it so difficult to have `@Before`/`@After` now, though?), but please open an issue to untangle this mess.

fanminshi:
@heyitsanthony I agree with you. I don't like how the tests are dependent on each other. I believe that the test is construct this way because all tests uses one etcd cluster. so ordering of tests that operate on the etcd cluster matters. I think if we have a test framework that each test case to have its independent etcd cluster. then we can make the test much more self contained. That's a TODO for me.

PR #108, opened 2017-06-02
maintenance: statusMember() and defragmentMember() now takes a endpoint

fanminshi:


fanminshi:
Work in Progress.

src/main/java/com/coreos/jetcd/MaintenanceImpl.java
```diff
@@ -38,10 +43,34 @@
...
+        .configureStub(MaintenanceGrpc.newFutureStub(c.getChannel()), c.getToken());
+    this.streamStub = ClientUtil
+        .configureStub(MaintenanceGrpc.newStub(c.getChannel()), c.getToken());
+  }
+
+  @FunctionalInterface
+  private interface DialFunction {
+
+    /**
+     * dial dials to an endpoint and returns a managed channel its associated token.
```
xiang90: a managed channel and its associated token

src/main/java/com/coreos/jetcd/MaintenanceImpl.java
```diff
@@ -38,10 +43,34 @@
...
+     *
+     * @return a managed channel its associated token.
```
xiang90: and its associated token

xiang90: i am not familiar with gRPC java. so a stupid question, why the token is needed?

src/main/java/com/coreos/jetcd/Client.java
```diff
@@ -45,30 +63,41 @@ public Client(ManagedChannelBuilder<?> channelBuilder, ClientBuilder clientBuild
...
 
+  Pair<ManagedChannel, Optional<String>> toChannelAndToken(ManagedChannelBuilder<?> channelBuilder)
+      throws AuthFailedException, ConnectException {
+    checkNotNull(channelBuilder, "channelBuilder can't be null");
+    ManagedChannel managedChannel = channelBuilder.build();
+    Optional<String> token = this.generateToken(managedChannel, this.user, this.pass);
+    return new Pair<>(managedChannel, token);
+  }
+
+  // ************************
```
xiang90: remove this line.

src/main/java/com/coreos/jetcd/MaintenanceImpl.java
xiang90: is this the authToken?

fanminshi: the token here is same as the one in go client counter part at https://github.com/coreos/etcd/blob/master/clientv3/client.go#L286.  If auth is enabled for etcd, the this token is needed to build a stub that has the token as its metadata. then this token is passed to etcd server for authentication.

src/main/java/com/coreos/jetcd/data/Pair.java
```diff
@@ -0,0 +1,34 @@
+package com.coreos.jetcd.data;
```
xiang90: not sure if this belongs to here. this is not actually etcd kv pair but some internal helper structure, right? i feel it is confusing to put it under data.

xiang90:
LGTM

src/main/java/com/coreos/jetcd/data/Pair.java
fanminshi: agreed. pair is merely a helper structure. I probaly have another pkg called `com.coreos.jetcd.internal` and put pair there?

xiang90: sgtm

PR #109, opened 2017-06-05
api: convert ListenableFuture to CompletableFuture

### LINKED ISSUE #16: Client interfaces

fanminshi:
FIXES #16 #46

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,6 +29,7 @@
         <!-- dependencies -->
         <grpc.version>1.2.0</grpc.version>
         <slf4j.version>1.7.21</slf4j.version>
+        <future-converter.version>1.1.0</future-converter.version>
 
         <!-- should be in sync with protobuf grpc dependencies -->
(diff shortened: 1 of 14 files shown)


xiang90:
lgtm

PR #110, opened 2017-06-06
Lease: Implement keepAliveOnce

fanminshi:


src/test/java/com/coreos/jetcd/LeaseTest.java
```diff
@@ -20,45 +21,58 @@
...
-    test.assertEquals(kvClient.get(testKey).get().getCount(), 0);
+    test.assertEquals(kvClient.get(KEY).get().getCount(), 0);
   }
 
   @Test(dependsOnMethods = "testRevoke")
+  public void testKeepAliveOnce() throws ExecutionException, InterruptedException {
+    long leaseID = leaseClient.grant(2).get().getID();
+    kvClient.put(KEY, VALUE, PutOption.newBuilder().withLeaseId(leaseID).build()).get();
+    test.assertEquals(kvClient.get(KEY).get().getCount(), 1);
+    Thread.sleep(1000);
```
xiang90: why do we need this sleep?

src/main/java/com/coreos/jetcd/LeaseImpl.java
```diff
@@ -216,16 +216,31 @@ public synchronized void cancelKeepAlive(long leaseId)
...
+          @Override
+          public void onCompleted() {
+          }
+        });
     requestObserver.onNext(newKeepAliveRequest(leaseId));
-    requestObserver.onCompleted();
 
-    throw new UnsupportedOperationException();
+    // cancel grpc stream when leaseKeepAliveResponseCompletableFuture completes.
+    // TODO: manage Async threads.
```
xiang90: what do you mean by managing Async threads?

src/test/java/com/coreos/jetcd/LeaseTest.java
```diff
@@ -20,45 +21,58 @@
...
+    leaseClient.keepAliveOnce(leaseID).get();
```
xiang90: check keep alive response here.

src/test/java/com/coreos/jetcd/LeaseTest.java
```diff
@@ -20,45 +21,58 @@
...
+    Thread.sleep(1500);
+    test.assertEquals(kvClient.get(KEY).get().getCount(), 1);
```
xiang90: remove 64-67. when we know server returns good response, we know keep alive request sent out correctly. we do not need to test server behavior in client test.

xiang90:
lgtm

src/main/java/com/coreos/jetcd/LeaseImpl.java
fanminshi: I think we should have an internal executor to handle the async threads. So when we close the client, we know where all our threads all. By default, the whenCompleteAsync without a exectuor use a common thread pool. 

"All async methods without an explicit Executor argument are performed using the ForkJoinPool.commonPool()" 
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html

src/test/java/com/coreos/jetcd/LeaseTest.java
fanminshi: sounds good.

src/test/java/com/coreos/jetcd/LeaseTest.java
fanminshi: probably don't need. we can just check  the  keepalive response instead.

src/main/java/com/coreos/jetcd/LeaseImpl.java
xiang90: ok, makes sense. need a better comment.

PR #111, opened 2017-06-07
client: replace ByteString with ByteSequence

fanminshi:
exposed APIs should take ByteSequence instead of ByteString as the argument type.
In this way, jetcd APIs doesn't depend on Guava ByteString.


xiang90:
lgtm after CI is fixed.

PR #115, opened 2017-06-12
Lease: rework keepAlive() api

fanminshi:


fanminshi:
Will add tests if the interface seems reasonable.

src/main/java/com/coreos/jetcd/Lease.java
```diff
@@ -50,57 +33,37 @@
...
-   * @return whether the keep alive service is running.
-   */
-  boolean isKeepAliveServiceRunning();
-
-  /**
-   * This interface is called by Etcd KeepAlive client to notify user about lease expiration and
-   * exception.
-   */
-  interface LeaseHandler {
+  interface KeepAliveListener {
```
xiang90: do we plan to use listener style thing for watcher too? i want the keepalive and watch api to be consistent. basically, we use the same method to notify user events. is it reasonable? /cc @heyitsanthony

fanminshi: For me, Watcher acts like a active agent. It watches stuff that changes. Listener is more a passive agent where it just receives whatever other tells it to do. In that sense, I am not sure if Listener style make sense for Watch; even though under the hood both just receives events from etcd server.

xiang90: for the watch api, we have two parts:

1. create watcher
2. get events from watcher

for part 1, it is different from keepalive where clients will send stuff. but for part2, it is similar when clients just listen to events, right?

src/main/java/com/coreos/jetcd/lease/KeepAlive.java
```diff
@@ -1,65 +1,72 @@
...
-  public KeepAlive(long leaseID) {
-    this(leaseID, null);
-  }
+  private DoneHandler doneHandler;
 
-  public KeepAlive(long leaseID, LeaseHandler leaseHandler) {
-    this.leaseID = leaseID;
-    this.leaseHandler = leaseHandler;
+  @FunctionalInterface
+  public interface DoneHandler {
```
heyitsanthony: It's not clear to me how this is useful; if the keepalive is closed by the user then the user already knows it's closing so there's no need for callback (or is close async? it probably shouldn't be...). If the connections close without user intervention, it's not obvious whether there were any error conditions.

src/main/java/com/coreos/jetcd/lease/KeepAlive.java
```diff
@@ -1,65 +1,72 @@
...
     return deadLine;
   }
 
-  public KeepAlive setDeadLine(long deadLine) {
+  public void setDeadLine(long deadLine) {
     this.deadLine = deadLine;
-    return this;
+  }
+
+  public void addListerner(KeepAliveListenerImpl listener) {
```
heyitsanthony: s/Listerner/Listener/

src/main/java/com/coreos/jetcd/Lease.java
heyitsanthony: yes, stream notification should be somewhat consistent

src/main/java/com/coreos/jetcd/lease/LeaseKeepAliveResponseWithError.java
```diff
@@ -0,0 +1,20 @@
...
+
+import com.coreos.jetcd.api.LeaseKeepAliveResponse;
+
+/**
+ * Created by fanminshi on 6/8/17.
+ */
+public class LeaseKeepAliveResponseWithError {
+
+  public LeaseKeepAliveResponse leaseKeepAliveResponse;
+  public Exception error;
```
heyitsanthony: is this thrown anywhere?

src/main/java/com/coreos/jetcd/Lease.java
fanminshi: @xiang90 that's correct. Maybe the Watcher should just have a `listen()` function instead of `take()`, `poll()`.  that seems cleaner.

xiang90: i am fine with listener.

src/main/java/com/coreos/jetcd/lease/KeepAlive.java
fanminshi: it is not async. I need a way tell LeaseImpl that this keepalive is done where all its listeners are closed by user. Then this keepalive needed be deleted from LeaseImpl's KeepAlives map. The handler is defined in LeaseImpl and all it does is to delete the keepalive.
```java
 @Override
  public synchronized KeepAliveListener keepAlive(long leaseId) {
...
    KeepAlive keepAlive = this.keepAlives.computeIfAbsent(leaseId, (key) -> {
      // the () -> this.removeKeepAlive(leaseId) is handler's definition
      KeepAlive ka = new KeepAlive(() -> this.removeKeepAlive(leaseId)); 
      long now = System.currentTimeMillis();
      ka.setDeadLine(now + FIRST_KEEPALIVE_TIMEOUT_MS);
      ka.setNextKeepAlive(now);
      return ka;
    });
...
```

src/main/java/com/coreos/jetcd/lease/LeaseKeepAliveResponseWithError.java
fanminshi: it is a poison pill that is used to notify the consumer thread something is wrong.
```java
@Override
    public synchronized LeaseKeepAliveResponse listen()
        throws InterruptedException {
...
      Future<LeaseKeepAliveResponse> future = service.submit(() -> {
        LeaseKeepAliveResponseWithError lkae = this.queue.take();
        if (lkae.error != null) {
          this.reason = lkae.error;
          throw lkae.error; // throw here. 
        }
        return lkae.leaseKeepAliveResponse;
      });
...
```

fanminshi:
added unit tests.

xiang90:
interface seems ok to me.

/cc @jlamillan @adohe @lburgazzoli can you help to review the implementation? Thanks!

src/main/java/com/coreos/jetcd/lease/LeaseKeepAliveResponseWithError.java
xiang90: i  guess the question is do we need to make this a exception type here?

src/main/java/com/coreos/jetcd/Util.java
```diff
@@ -75,8 +76,15 @@ protected static WatchEvent apiToClientEvent(Event event) {
   /**
    * convert API response header to self defined header.
    */
-  protected static Header apiToClientHeader(ResponseHeader header, long compactRevision) {
+  static Header apiToClientHeader(ResponseHeader header, long compactRevision) {
```
xiang90: just ToHeader? why do we add the api prefix?

src/main/java/com/coreos/jetcd/Util.java
```diff
@@ -75,8 +76,15 @@ protected static WatchEvent apiToClientEvent(Event event) {
...
     return new Header(header.getClusterId(), header.getMemberId(), header.getRevision(),
         header.getRaftTerm(), compactRevision);
   }
+
+  static com.coreos.jetcd.lease.LeaseKeepAliveResponse apiToLeaseKeepAliveResponse(
```
xiang90: why api prefix?

src/main/java/com/coreos/jetcd/LeaseImpl.java
```diff
@@ -62,158 +65,192 @@
    */
   private StreamObserver<LeaseKeepAliveResponse> keepAliveResponseStreamObserver;
 
+  private boolean firstKeepAlive = true;
```
xiang90: firstKeepAliveSeen = false? this boolean member variable is not super descriptive.

src/main/java/com/coreos/jetcd/LeaseImpl.java
```diff
@@ -62,158 +65,192 @@
...
+      LeaseKeepAliveResponse leaseKeepAliveResponse) {
+    if (this.closed) {
+      return;
+    }
+
+    long leaseID = leaseKeepAliveResponse.getID();
+    long ttl = leaseKeepAliveResponse.getTTL();
+
+    KeepAlive ka = this.keepAlives.get(leaseID);
+    if (ka == null) {
```
xiang90: when will this happen? add a comment?

src/main/java/com/coreos/jetcd/LeaseImpl.java
```diff
@@ -62,158 +65,192 @@
...
+      return;
+    }
+
+    if (ttl <= 0) {
+      // lease expired; close all keep alive
+      this.removeKeepAlive(leaseID);
+      ka.sentKeepAliveResp(new LeaseKeepAliveResponseWithError(
+          new IllegalStateException("Lease " + leaseID + " expired")));
```
xiang90: not sure this is an IllegalState.

src/main/java/com/coreos/jetcd/lease/KeepAlive.java
```diff
@@ -1,65 +1,72 @@
...
   private long nextKeepAlive;
 
-  private LeaseHandler leaseHandler;
+  // ownerLock protects owner map.
+  private final Object ownerLock;
 
-  public KeepAlive(long leaseID) {
-    this(leaseID, null);
-  }
+  private Map<Long, KeepAlive> owner;
```
xiang90: does not jave have concurrent map or something?

src/main/java/com/coreos/jetcd/Util.java
fanminshi: I think the original author want to clarify that the conversion is between API header to Custom Header. 
I am fine with remove the API prefix.

xiang90:
The implementation looks reasonable in general. I am not too worry about making it "perfect" for now.

src/main/java/com/coreos/jetcd/LeaseImpl.java
fanminshi: yeah, i'll rework on this one.

xiang90: or just started? not sure.

src/main/java/com/coreos/jetcd/lease/KeepAlive.java
fanminshi: the concurrentHashMap is passed in from leaseImpl through its constructor. So it will hold a reference to a concurrent hashmap.

xiang90: ah. i will give this a more careful look later today or tomorrow.

src/main/java/com/coreos/jetcd/LeaseImpl.java
fanminshi: I need to rework the exception. right now it is just a placeholder for unchecked exception.
I took a look of spanner api. they just have one SpannerException with error code and error msg. 
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerException.java

I probably will copy that structure but i need to think through how to handle with error code.

xiang90: ok. i think etcdException is the way to go in the future.

fanminshi: yeah, I'll need to put more thought into this.

src/main/java/com/coreos/jetcd/LeaseImpl.java
fanminshi: When ka is closed due all the listeners are closed just before a ka response for the lease id arrives. In that case, don't broadcast ka response if the ka is deleted.

fanminshi:
@xiang90 I'll leave this open for other to review. In the meantime, I'll think about exception handling.

xiang90:
@fanminshi Let's try to clean this up, resolve conflicts, and get merged this week.

fanminshi:
@xiang90 will do.

fanminshi:
@xiang90 done resolving conflicts.

xiang90:
looks ok in general. we can iterate on the impl in the future.

fanminshi:
yeah, i think next step is add better exception handling.

PR #119, opened 2017-06-15
Lease: add timeToLive() api

fanminshi:


src/main/java/com/coreos/jetcd/LeaseImpl.java
```diff
@@ -238,11 +245,27 @@ public void onCompleted() {
...
+    lkaFuture
+        .whenCompleteAsync((val, throwable) -> requestObserver.onCompleted(), this.executorService);
 
     return lkaFuture;
   }
 
+  @Override
+  public CompletableFuture<com.coreos.jetcd.lease.LeaseTimeToLiveResponse> timeToLive(long leaseId,
+      LeaseOption option) {
+    checkNotNull(option, "LeaseOption should not be null");
```
xiang90: maybe null is OK?

src/main/java/com/coreos/jetcd/Util.java
```diff
@@ -79,4 +84,49 @@ protected static Header apiToClientHeader(ResponseHeader header, long compactRev
     return new Header(header.getClusterId(), header.getMemberId(), header.getRevision(),
         header.getRaftTerm(), compactRevision);
   }
+
+  /**
+   * convert API LeaseTimeToLiveResponse to self defined LeaseTimeToLiveResponse.
+   */
+  static LeaseTimeToLiveResponse apiToLeaseTimeToLiveResponse(
```
xiang90: remove api prefix?

src/main/java/com/coreos/jetcd/Util.java
```diff
@@ -79,4 +84,49 @@ protected static Header apiToClientHeader(ResponseHeader header, long compactRev
...
+        .collect(Collectors.toList());
+    return new LeaseTimeToLiveResponse(apiToClientHeader(response.getHeader(), 0),
+        response.getID(), response.getTTL(), response.getGrantedTTL(), byteSequenceKeys);
+  }
+
+  /**
+   * convert ListenableFuture of Type S to CompletableFuture of Type T. copied from:
+   * https://github.com/enncloud/jetcd/blob/master/src/main/java/com/coreos/jetcd/EtcdUtil.java
+   */
+  static <S, T> CompletableFuture<T> listenableToCompletableFuture(
```
xiang90: why do we need the copy? can we use the one in EtcdUtil.java?

src/main/java/com/coreos/jetcd/lease/LeaseTimeToLiveResponse.java
```diff
@@ -0,0 +1,79 @@
...
+      return false;
+    }
+    if (!getHeader().equals(that.getHeader())) {
+      return false;
+    }
+    return getKeys().equals(that.getKeys());
+  }
+
+  @Override
+  public int hashCode() {
```
xiang90: why do we need this?

xiang90:
looks good in general

src/main/java/com/coreos/jetcd/Util.java
fanminshi: will do that.

src/main/java/com/coreos/jetcd/Util.java
fanminshi: it is not from our branch. it is from enncloud's `https://github.com/enncloud/jetcd/`. So i copied his code.

xiang90: i thought we have this somewhere in our codebase?

src/main/java/com/coreos/jetcd/LeaseImpl.java
fanminshi: I saw the kvImpl doesn't allow `null`l option. so I follow that. I wonder which one seems more correct?
`timeToLive(leaseID, LeaseOption.DEFAULT)` or`timeToLive(leaseID, null)`

xiang90: what is the java practice? what do other projects do?

src/main/java/com/coreos/jetcd/lease/LeaseTimeToLiveResponse.java
fanminshi: generally it is a good idea to override equals and hashcode so that two objects are equal not because of the their reference but because of the contents. So put those same objects in hash map or set only result one space taken  instead of many. 

some explanation:
http://www.ideyatech.com/effective-java-equals-and-hashcode/

src/main/java/com/coreos/jetcd/Util.java
fanminshi: it wasn't merged in. so we don't have it.

src/main/java/com/coreos/jetcd/LeaseImpl.java
fanminshi: from here `Avoid null if possible`

http://www.javapractices.com/topic/TopicAction.do?Id=134

from CleanCode:
it says Don't Pass Null.

src/main/java/com/coreos/jetcd/Util.java
fanminshi: I can cherry pick that specific commit from enncloud before we merge this one.

src/main/java/com/coreos/jetcd/lease/LeaseTimeToLiveResponse.java
xiang90: i understand why it is good in some cases to overwrite this. But i feel we do not really need this in our case? less code == less to maintain and worry about.

src/main/java/com/coreos/jetcd/LeaseImpl.java
xiang90: ok

src/main/java/com/coreos/jetcd/Util.java
xiang90: probably just create the util pkg like what it does.

src/main/java/com/coreos/jetcd/lease/LeaseTimeToLiveResponse.java
fanminshi: sounds good. I'll remove it then

src/main/java/com/coreos/jetcd/Util.java
fanminshi: sorry I didn't get what do you mean by `create the util pkg like what it does.`? there is util class already. Do you mean create a Util pkg like `com.coreos.jetcd.util` and somehow put `listenableToCompletableFuture()` wrapped in its own class there?

xiang90: put this func here.

remove comment 

> copied from https://github.com/enncloud/jetcd/blob/master/src/main/java/com/coreos/jetcd/EtcdUtil.java

xiang90:
lgtm

src/main/java/com/coreos/jetcd/Util.java
fanminshi: got it.

PR #120, opened 2017-06-16
Lease: refactor lease APIs to take client side response

fanminshi:
hide grpc response from user by return client side response.
use listenableToCompletableFuture() instead of FutureConverter.toCompletableFuture() for convert listenable future.
minor refactoring on util and watch


xiang90:
lgtm

PR #121, opened 2017-06-17
kv: add comments to KV APIs

fanminshi:


fanminshi:
prelim pr for replacing grpc response to client response for KV.

fanminshi:
however, this pr might make api more verbose to use.

fanminshi:
removing api wrapper seems too verbose. I'll add them back.

fanminshi:
prelim pr to replace grpc response with client kv response.

src/main/java/com/coreos/jetcd/KV.java
```diff
@@ -20,36 +20,72 @@
...
+   * put a key-value pair into etcd.
+   *
+   * @param key key in ByteSequence
+   * @param value value in ByteSequence
+   * @return PutResponse
+   */
   CompletableFuture<PutResponse> put(ByteSequence key, ByteSequence value);
 
+  /**
+   * put a key-value pair into etcd.
```
xiang90: put a key-value pair into etcd with option.

src/main/java/com/coreos/jetcd/KV.java
```diff
@@ -20,36 +20,72 @@
...
+   * @return PutResponse
+   */
   CompletableFuture<PutResponse> put(ByteSequence key, ByteSequence value, PutOption option);
 
-  // ***************
-  // Op.GET
-  // ***************
-
+  /**
+   * retrieve keys. By default, get will return the value for "key".
```
xiang90: i do not understand what does the second sentence imply.

src/main/java/com/coreos/jetcd/KV.java
```diff
@@ -20,36 +20,72 @@
...
+  /**
+   * delete a key.
+   *
+   * @param key key in ByteSequence
+   * @return DeleteRangeResponse
+   */
   CompletableFuture<DeleteRangeResponse> delete(ByteSequence key);
 
+  /**
+   * delete a key.
```
xiang90: delete a key or range with option.

src/main/java/com/coreos/jetcd/KV.java
```diff
@@ -20,36 +20,72 @@
...
-  // ***************
-  // Op.COMPACT
-  // ***************
-
+  /**
+   * compact etcd KV history before the given rev.
+   *
+   * @return CompactionResponse
+   */
   CompletableFuture<CompactionResponse> compact();
```
xiang90: hmm... why do we need compact without option? what does go client do?

xiang90:
lgtm

src/main/java/com/coreos/jetcd/KV.java
fanminshi: I copied it from go client. I think it means that without option, the get just return value that's corresponding to the key.

```go
	// Get retrieves keys.
	// By default, Get will return the value for "key", if any.
	// When passed WithRange(end), Get will return the keys in the range [key, end).
	// When passed WithFromKey(), Get returns keys greater than or equal to key.
	// When passed WithRev(rev) with rev > 0, Get retrieves keys at the given revision;
	// if the required revision is compacted, the request will fail with ErrCompacted .
	// When passed WithLimit(limit), the number of returned keys is bounded by limit.
	// When passed WithSort(), the keys will be sorted.
	Get(ctx context.Context, key string, opts ...OpOption) (*GetResponse, error)
```

https://github.com/coreos/etcd/blob/master/clientv3/kv.go#L38-L39

src/main/java/com/coreos/jetcd/KV.java
fanminshi: good catch. the compact API should take at least one arg which is the rev to be compacted at. I'll fix this.

src/main/java/com/coreos/jetcd/KV.java
xiang90: that makes sense since go client get has an option field. for this func, we only have one arg, there is no option just "default".

fanminshi: I'll just change the comment to `retrieve value of a given key`.

PR #122, opened 2017-06-19
KV: KV returns client side response and takes in client as the constructor arg

fanminshi:
remove KV grpc dependency on PutResponse, RangeResponse, DeleteRangeResponse, and CompactionResponse.

KV constructor takes in client and uses client's executor, channel to build KVImpl.


xiang90:
lgtm

PR #123, opened 2017-06-20
set WatchID first and then put to watchers when creating a watch

WlaeniG:


fanminshi:
please change to commit msg to 
`watch: set WatchID first and then put to watchers when creating a watch.`
thanks.

fanminshi:
once you changed the commit msg. then lgtm.

fanminshi:
close via #137

PR #127, opened 2017-06-22
proto: update rpc.proto at etcd f7df3c80d5d9fe4782cfb28c01a3c9bf766a15aa

fanminshi:
updated rpc.proto allows java client to implement nested txn and retrieve cancel_reason field from watch response.


xiang90:
lgtm

PR #128, opened 2017-06-22
test: don't run testKeepAliveOnceStreamCloseOnSuccess in LeaseUnitTest

fanminshi:
it seems like there is a bug in grpc test framework that's blocking on onNext().


heyitsanthony:
can this bug be confirmed so it can be fixed?

fanminshi:
@heyitsanthony I'll open another issue to track this bug.

Issue #116, opened 2017-06-15
[discuss] API implementation visibility/packages


lburgazzoli:
The package com.coreos.jetcd contains both the interface definition and the related implementations that cane be instantiated as they are public and have a public constructor so I think it would be nice to do eithe make the visibility of the implemntation classes as package scope or move them to com.coreos.jetcd.internal.impl so at least the name mark them as non public.

Thoughts ?

fanminshi:
I am fine with moving impl classes to `com.coreos.jetcd.internal.impl` folder.

xiang90:
@lburgazzoli is this already fixed?

lburgazzoli:
yes it is

PR #132, opened 2017-06-23
API implementation visibility/packages #116

lburgazzoli:


src/main/java/com/coreos/jetcd/internal/impl/Util.java
```diff
@@ -163,7 +164,7 @@ static DeleteResponse toDeleteResponse(DeleteRangeResponse rangeResponse) {
         toKVs(rangeResponse.getPrevKvsList()));
   }
 
-  private static List<KeyValue> toKVs(List<com.coreos.jetcd.api.KeyValue> kvs) {
+  public  static List<KeyValue> toKVs(List<com.coreos.jetcd.api.KeyValue> kvs) {
```
fanminshi: why public? is `toKVs` used anywhere else besides Util pkg?

fanminshi:
can you change the commit title to be
`*:API implementation visibility/packages #116`
and lgtm.

xiang90:
lgtm. merging.

@lburgazzoli thanks!

PR #135, opened 2017-06-27
WatchOption add withPrefix for watch a directory

donsonHub:
WatchOption add withPrefix for watch a directory

PR #138, opened 2017-06-28
options: add withPrefix in WatchOption

fanminshi:


fanminshi:
same as #135 but with commit title changed to that agree with etcd commit style.

PR #140, opened 2017-06-28
watch: add client side WatchResponse and WatchResponseWithError

fanminshi:
this is a prelim pr for watch rewrite.

### DIFF
--- a/src/main/java/com/coreos/jetcd/watch/WatchResponse.java
+++ b/src/main/java/com/coreos/jetcd/watch/WatchResponse.java
@@ -1,0 +1,23 @@
+package com.coreos.jetcd.watch;
+
+import com.coreos.jetcd.data.Header;
+import java.util.List;
+
+public class WatchResponse {
+
+  private Header header;
+  private List<WatchEvent> events;
+
+  public WatchResponse(Header header, List<WatchEvent> events) {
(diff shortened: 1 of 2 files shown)


fanminshi:
#139 is merged and rebased this pr on top of that. fixed the ci issue.

src/main/java/com/coreos/jetcd/watch/WatchResponse.java
```diff
@@ -0,0 +1,23 @@
+package com.coreos.jetcd.watch;
+
+import com.coreos.jetcd.data.Header;
+import java.util.List;
+
+public class WatchResponse {
+
+  private Header header;
+  private List<WatchEvent> events;
```
xiang90: is watcheven defined by our client library already?

fanminshi: it is defined under `com.coreos.jetcd.watch` pkg.

fanminshi: https://github.com/coreos/jetcd/blob/master/src/main/java/com/coreos/jetcd/watch/WatchEvent.java

xiang90:
lgtm

PR #151, opened 2017-06-29
api: redefines Txn API and hides Txn Grpc Response

fanminshi:


src/main/java/com/coreos/jetcd/Txn.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+public interface Txn {
+
+  /**
+   * takes a list of comparison. If all comparisons passed in succeed,
+   * the operations passed into Then() will be executed. Or the operations
+   * passed into Else() will be executed.
+   */
+  //CHECKSTYLE:OFF
+  Txn If(Cmp... cmps);
```
xiang90: if ?

src/main/java/com/coreos/jetcd/Txn.java
```diff
@@ -0,0 +1,77 @@
...
+  //CHECKSTYLE:ON
+
+  /**
+   * takes a list of operations. The Ops list will be executed, if the
+   * comparisons passed in If() succeed.
+   */
+  //CHECKSTYLE:OFF
+  Txn Then(Op... ops);
```
xiang90: then?

src/main/java/com/coreos/jetcd/Txn.java
```diff
@@ -0,0 +1,77 @@
...
+  //CHECKSTYLE:ON
+
+  /**
+   * takes a list of operations. The Ops list will be executed, if the
+   * comparisons passed in If() fail.
+   */
+  //CHECKSTYLE:OFF
+  Txn Else(Op... ops);
```
xiang90: else?

src/main/java/com/coreos/jetcd/kv/TxnResponse.java
```diff
@@ -0,0 +1,37 @@
+package com.coreos.jetcd.kv;
+
+import com.coreos.jetcd.data.Header;
+import java.util.List;
+
+public class TxnResponse {
```
xiang90: doc string?

src/main/java/com/coreos/jetcd/kv/TxnResponse.java
```diff
@@ -0,0 +1,37 @@
...
+    this.deleteResponses = deleteResponses;
+    this.getResponses = getResponses;
+    this.putResponses = putResponses;
+  }
+
+  public Header getHeader() {
+    return header;
+  }
+
+  public List<DeleteResponse> getDeleteResponses() {
```
xiang90: doc string?

src/main/java/com/coreos/jetcd/kv/TxnResponse.java
```diff
@@ -0,0 +1,37 @@
...
+    return deleteResponses;
+  }
+
+  public List<GetResponse> getGetResponses() {
```
xiang90: doc string?

src/main/java/com/coreos/jetcd/kv/TxnResponse.java
```diff
@@ -0,0 +1,37 @@
...
+    return getResponses;
+  }
+
+  public List<PutResponse> getPutResponses() {
```
xiang90: doc string?

xiang90:
lgtm

src/main/java/com/coreos/jetcd/Txn.java
fanminshi: `if` is a reserved key word in java for if statement.  can't use that for function name.

src/main/java/com/coreos/jetcd/Txn.java
fanminshi: `else` is also reserved. can't use that.

src/main/java/com/coreos/jetcd/kv/TxnResponse.java
fanminshi: will do

src/main/java/com/coreos/jetcd/kv/TxnResponse.java
fanminshi: will do.

src/main/java/com/coreos/jetcd/kv/TxnResponse.java
fanminshi: will do

src/main/java/com/coreos/jetcd/Txn.java
xiang90: oh... i see. keep if :P

src/main/java/com/coreos/jetcd/kv/TxnResponse.java
fanminshi: will do

src/main/java/com/coreos/jetcd/Txn.java
xiang90: ack

PR #152, opened 2017-06-30
Update etcd version

lburgazzoli:


scripts/run_etcd.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-ETCD_VERSION="v3.1.8"
```
fanminshi: let's update to 3.2.1
https://github.com/coreos/etcd/releases/tag/v3.2.1

fanminshi:
then lgtm

fanminshi:
also can you change commit title to be
`version: update etcd version to 3.2.1` thanks.

PR #156, opened 2017-07-06
op: add client endpoint format check

joshua-jin:


src/main/java/com/coreos/jetcd/internal/impl/ClientUtil.java
```diff
@@ -63,4 +66,36 @@ static URI endpointToUri(String endpoint) {
...
+    return urlValidator.isValid(url);
+  }
+
+  private static boolean isValidHost(String host) {
+    return InetAddressValidator.getInstance().isValid(host) || "localhost".equals(host);
+  }
+
+  private static boolean isValidPort(String port) {
+    IntegerValidator portValidator = IntegerValidator.getInstance();
+    return portValidator.isValid(port) && portValidator.isInRange(Integer.parseInt(port), 0, 65535);
```
ScienJus: maybe we should handle `NumberFormatException` ?

src/main/java/com/coreos/jetcd/internal/impl/ClientUtil.java
```diff
@@ -63,4 +66,36 @@ static URI endpointToUri(String endpoint) {
...
+
+    String host = hostAndPort[0];
+    String port = hostAndPort[1];
+
+    return (isValidHost(host) && isValidPort(port));
+  }
+
+  private static boolean isValidURL(String url) {
+    String[] scheme = {"http"};
+    UrlValidator urlValidator = new UrlValidator(scheme, UrlValidator.ALLOW_LOCAL_URLS);
```
ScienJus: i think `urlValidator` should be a singleton instance.

joshua-jin: Can't change default schemes and options, if use `UrlValidator .getInstance()` method.
Tell me if I misunderstand something.

ScienJus: maybe just need a static field in `ClientUtils` instead of creating for each execution.

fanminshi:
```java
  private final static List<String> SCHEMES = Arrays.asList("http", "https", "unix", "unixs");

  public static boolean isValidEndpointFormat(String endpoint) {
    URL u = null;
    try {
      u = new URL(endpoint);
    } catch (MalformedURLException e) {
      return false;
    }

    if (SCHEMES.stream().noneMatch(u.getProtocol()::equals)) {
      return false;
    }

    // endpoint must not contain a path.
    if (!u.getPath().isEmpty()) {
      return false;
    }

    return true;
  }
```

wouldn't above be sufficient to check validity of enpoint url?

src/test/java/com/coreos/jetcd/internal/impl/ClientBuilderTest.java
```diff
@@ -39,4 +48,32 @@ public void testEndPoints_Verify_SomeEmpty() {
...
+  public void testBuild_WithValidFormatEndpoints() {
+    builder.setEndpoints("127.0.0.1:2379", "http://localhost:1984", "http://www.foo.com:2300/",
+                         "http://192.168.1.10:8888/");
+
+    test.assertEquals(builder.getEndpoints().size(), 4);
+
+    test.assertTrue(builder.getEndpoints().contains("127.0.0.1:2379"));
+    test.assertTrue(builder.getEndpoints().contains("http://localhost:1984"));
+    test.assertTrue(builder.getEndpoints().contains("http://www.foo.com:2300/"));
+    test.assertTrue(builder.getEndpoints().contains("http://192.168.1.10:8888/"));
```
fanminshi: please add test for IPV6
`http://[fe80::6e40:8ff:fe]:2380")`

src/main/java/com/coreos/jetcd/internal/impl/ClientUtil.java
fanminshi:
```java
  private final static List<String> SCHEMES = Arrays.asList("http", "https", "unix", "unixs");

  public static boolean isValidEndpointFormat(String endpoint) {
    URL u = null;
    try {
      u = new URL(endpoint);
    } catch (MalformedURLException e) {
      System.out.println(e);
      return false;
    }

    if (SCHEMES.stream().noneMatch(u.getProtocol()::equals)) {
      return false;
    }

    // endpoint must contain a port.
    System.out.println(u.getPort());
    if(u.getPort() == -1){
      return false;
    }

    // endpoint must not contain a path.
    if (!u.getPath().isEmpty()) {
      return false;
    }

    return true;
  }
```
I think the above should be sufficient to check validity of a endpoint.

src/main/java/com/coreos/jetcd/ClientBuilder.java
```diff
@@ -122,7 +123,7 @@ public AbstractEtcdNameResolverFactory getNameResolverFactory() {
    * build a new Client.
    *
    * @return Client instance.
-   * @throws ConnectException As network reason, wrong address
+   * @throws ConnectException    As network reason, wrong address
```
fanminshi: no need to have extra space

src/main/java/com/coreos/jetcd/ClientBuilder.java
```diff
@@ -44,18 +45,18 @@ public static ClientBuilder newBuilder() {
    *
    * @param endpoints etcd server endpoints, at least one
    * @return this builder to train
-   * @throws NullPointerException if endpoints is null or one of endpoint is null
+   * @throws NullPointerException     if endpoints is null or one of endpoint is null
```
fanminshi: no need to have extra space.

pom.xml
```diff
@@ -72,6 +72,11 @@
             <artifactId>future-converter-java8-guava</artifactId>
             <version>${future-converter.version}</version>
         </dependency>
+        <dependency>
```
fanminshi: no need to import extra pkg. I think using URL pkg is enough for checking url.

src/test/java/com/coreos/jetcd/internal/impl/ClientBuilderTest.java
```diff
@@ -39,4 +48,32 @@ public void testEndPoints_Verify_SomeEmpty() {
...
+      builder.setEndpoints("xxx.yyy.111.222:8999");
+  }
+
+  @Test(expectedExceptions = IllegalArgumentException.class)
+  public void testBuild_WithInvalidPortEndpoint() {
+    builder.setEndpoints("222.255.111.222:uyw3");
+  }
+
+  @Test(expectedExceptions = IllegalArgumentException.class)
+  public void testBuild_WithoutPortEndpoint() {
```
fanminshi: you can aggregate invalidEndpoint testing using the following technique.
```java
  @Test(dataProvider = "invalidEndpoints", expectedExceptions = IllegalArgumentException.class)
  public void testBuildInvalidEndpointFormat(String endpoint) {
    builder.setEndpoints(endpoint);
  }

  @DataProvider
  Object[][] invalidEndpoints() {
    return new Object[][]{
        // missing protocol scheme
        {"://127.0.0.1:2379" },
        // unsupported scheme
        {"mailto://127.0.0.1:2379" },
        // not conform to host:port
        {"http://127.0.0.1" },
        // contain a path
        {"http://127.0.0.1:2379/path" },
        // more...
    };
  }
```

src/test/java/com/coreos/jetcd/internal/impl/ClientBuilderTest.java
```diff
@@ -39,4 +48,32 @@ public void testEndPoints_Verify_SomeEmpty() {
   public void testBuild_WithoutEndpoints() throws AuthFailedException, ConnectException {
     builder.build();
   }
+
+  @Test
+  public void testBuild_WithValidFormatEndpoints() {
```
fanminshi: you should use the `DataProvider` for this test case too.
see comment below.

fanminshi:
also please resolve merge conflicts. thanks.

src/main/java/com/coreos/jetcd/internal/impl/ClientUtil.java
joshua-jin: I have run this block of code, but it didn't accept `127.0.0.1:2379` as a valid endpoint

src/main/java/com/coreos/jetcd/ClientBuilder.java
```diff
@@ -165,4 +173,39 @@ public ClientBuilder copy() {
       throw EtcdExceptionFactory.newEtcdException(e);
     }
   }
+
+  private static boolean isValidEndpointFormat(String endpoint) {
+    if (isValidHostAndPort(endpoint)) {
```
fanminshi: why is this needed? valid host and port will be checked when URL are constructed.

src/main/java/com/coreos/jetcd/ClientBuilder.java
```diff
@@ -165,4 +173,39 @@ public ClientBuilder copy() {
...
+    // endpoint must contain a port.
+    if (u.getPort() == -1) {
+      return false;
+    }
+
+    // endpoint must not contain a path.
+    return u.getPath().isEmpty();
+  }
+
+  private static boolean isValidHostAndPort(String endpoint) {
```
fanminshi: no need to have this helper function.

fanminshi: nvm, i think `isValidHostAndPort` is used to validate `"localhost:2379"`, `"127.0.0.1:2379"`, and `"fe80::6e40:8ff:fe:2380"`

fanminshi: @xiang90 should client accept address in form of `"localhost:2379"` and `"127.0.0.1:2379"` without scheme specified?

xiang90: it should not. scheme should be specific. unless java client wants to attach http as default scheme.

fanminshi: @xiang90 agreed. I think endpoint specified should be specific and valid.

src/test/java/com/coreos/jetcd/internal/impl/ClientBuilderTest.java
```diff
@@ -37,4 +38,37 @@ public void testEndPoints_Verify_SomeEmpty() {
...
+  public void testBuild_WithValidFormatEndpoints(String endpoint) {
+    builder.setEndpoints(endpoint);
+  }
+
+  @DataProvider
+  private Object[][] validEndpoints() {
+    return new Object[][] {
+        {"localhost:2379"},
+        {"127.0.0.1:2379"},
+        {"fe80::6e40:8ff:fe:2380"},
```
fanminshi: don't let `"localhost:2379"`, `"127.0.0.1:2379"`, and `"fe80::6e40:8ff:fe:2380"` to be valid endpoint since they don't specify a scheme.

src/main/java/com/coreos/jetcd/ClientBuilder.java
fanminshi: no need to have this.

src/main/java/com/coreos/jetcd/ClientBuilder.java
joshua-jin: @xiang90 so you mean `ClientBuilder.setEndpoints` method should always accept endpoint string with scheme ?

fanminshi: @joshua-jin I let's just always accept endpoint with scheme first. we can iterate on this later if user want the un scheme version.

fanminshi:
please resolve the merge conflict. Then lgtm.

fanminshi:
@joshua-jin don't include the `merge: merge with latest master`.  this commit shouldn't include anything relate other commit.

to resolve merge conflict. you should do the following.
```sh
# update local master
$ git checkout master
$ git fetch upstream
$ git merge upstream/master

# resolve merge conflict with master via rebase
$ git checkout endpoint-check
$ git rebase master
...
```

src/main/java/com/coreos/jetcd/internal/impl/ClientUtil.java
```diff
@@ -0,0 +1,132 @@
+package com.coreos.jetcd.internal.impl;
```
joshua-jin: to resolve conflict of `SimpleEtcdNameResolverFactory`

fanminshi: I think you might resolve merge conflict in-correctly. your pr doesn't use `SimpleEtcdNameResolverFactory` anywhere.

When you rebase your branch on master. Just remove ClientUtil.java since it is not using anywhere.

fanminshi: it is probably the easiest to create a new pr and add the changes on top of the master branch; probably faster than resolve the merge conflict of this pr.

joshua-jin: ok, I will close this pr and create a new one to solve this problem

fanminshi: sounds good!

PR #161, opened 2017-07-11
Make it easy to configure ssl

lburgazzoli:
This PR lacks of tests, any help on properly setting up an
etcd instance with ssl would be really welcome


src/main/java/com/coreos/jetcd/ClientBuilder.java
```diff
@@ -145,6 +148,20 @@ public ClientBuilder setLazyInitialization(boolean lazyInitialization) {
...
+    return sslContext;
+  }
+
+  /**
+   * SSL/TLS context to use instead of the system default. It must have been configured with {@link
+   * GrpcSslContexts}, but options could have been overridden.
+   *
+   * @param sslContext the ssl context
+   */
+  public void setSslContext(SslContext sslContext) {
```
xiang90: java still prefer ssl than tls?

fanminshi:
@lburgazzoli the current integration test platform is not good. It needs to be worked on. In the meantime, could you some sort of manual test? And then post the manual test result to this pr.

src/main/java/com/coreos/jetcd/resolver/SimpleEtcdNameResolverFactory.java
```diff
@@ -46,4 +49,24 @@ public String getDefaultScheme() {
   public String name() {
     return NAME;
   }
+
+  public static SimpleEtcdNameResolverFactory forEndpoints(String... endpoints) {
```
fanminshi: prefer not to have `Etcd` keyword in naming unless you think that it's more clear.

src/main/java/com/coreos/jetcd/resolver/SimpleEtcdNameResolverFactory.java
```diff
@@ -46,4 +49,24 @@ public String getDefaultScheme() {
...
+    return forEndpoints(Arrays.asList(endpoints));
+  }
+
+  public static SimpleEtcdNameResolverFactory forEndpoints(List<String> endpoints) {
```
fanminshi: prefer not to have `Etcd` keyword in naming unless you think that it's more clear.

src/main/java/com/coreos/jetcd/ClientBuilder.java
fanminshi: @xiang90 it seems so. if you look the grpc/auth guide for java; "ssl" is used instead of "tls" in naming. https://grpc.io/docs/guides/auth.html#with-server-authentication-ssltls-4

src/main/java/com/coreos/jetcd/resolver/SimpleEtcdNameResolverFactory.java
lburgazzoli: I will change it but I'm not the author of the class :)

fanminshi: oh i see. My bad. don't change it. probably do it in a different pr.

src/main/java/com/coreos/jetcd/resolver/SimpleEtcdNameResolverFactory.java
fanminshi: nvm.

lburgazzoli:
I will try to add some tests as soon as possible but in the meantime it would be nice to merge this PR so people can eventually test it and report issues.

fanminshi:
@lburgazzoli still reviewing the code; need to understand how tls works in grpc.

lburgazzoli:
In the meantime I've renamed the class, not an issue to change it in this PR IMHO as lots of things are still moving ;)

src/main/java/com/coreos/jetcd/ClientBuilder.java
xiang90: ok. good to know.

xiang90:
can we add a test for this? or please create an issue to track ssl testing.

lburgazzoli:
@xiang90 I've opened the issue #162

src/main/java/com/coreos/jetcd/resolver/SimpleNameResolverFactory.java
```diff
@@ -46,4 +49,24 @@ public String getDefaultScheme() {
...
+          endpoint = "http://" + endpoint;
+        }
+        return new URI(endpoint);
+      } catch (URISyntaxException e) {
+        throw new IllegalArgumentException(e);
+      }
+    }).collect(Collectors.toList());
+
+    return new SimpleNameResolverFactory(uris);
+
```
fanminshi: no extra space after return?

src/main/java/com/coreos/jetcd/internal/impl/ClientConnectionManager.java
```diff
@@ -136,4 +146,87 @@ synchronized void close() {
...
+    channelBuilder.nameResolverFactory(factory);
+
+    if (builder.getSslContext() != null) {
+      channelBuilder.sslContext(builder.getSslContext());
+    } else {
+      channelBuilder.usePlaintext(true);
+    }
+
+    return channelBuilder;
+
```
fanminshi: no extra space after return?

pom.xml
```diff
@@ -109,6 +110,13 @@
           <version>${grpc.version}</version>
           <scope>test</scope>
         </dependency>
+        <dependency>
+          <groupId>io.netty</groupId>
+          <artifactId>netty-tcnative-boringssl-static</artifactId>
+          <version>${netty-tcnative.version}</version>
+          <classifier>${os.detected.classifier}</classifier>
+          <scope>test</scope>
```
lburgazzoli: Is is indeed for test only as jetcd user should be able to use different versions (i.e. dynamic linked) so netty-tcnative should not be a transitive dependency

src/main/java/com/coreos/jetcd/internal/impl/ClientConnectionManager.java
lburgazzoli: I'd recommend to improve checkstyle to include any style check you'd need.

pom.xml
fanminshi: Suppose that I export jetcd as jetcd.jar. If the user wants to use jetcd.jar, then the user also needs include a ` netty-tcnative` pkg in order to build the code with jetcd.jar properly? since jetcd needs to be dynamically linked to a `netty-tcnative` pkg. Is my understand of `dynamic linked` in for java jar correct?

lburgazzoli: no netty-tcnatyve is not required to run jetcd, it is an optional dependency if you want to use a fast ssl implementation based on openssl.

fanminshi: Sorry I am not an export on maven dependency. More question, If I scope the `netty-tcnative-boringssl-static` to test. then the compiled scope of jetcd will not have `io.netty.handler.ssl.SslContext` defined anywhere. this means that the user must import another jar that has `io.netty.handler.ssl.SslContext`?

lburgazzoli: SslContext is not part of netty-tcnative which contains only the native binding (via jni) for openssl so the implementation to use is chosen at runtime according to the classpath.

fanminshi: I think I know what's going on. `Netty transport` used by jetcd's grpc needs a mechanism to do TLS. The `netty-tcnative-boringssl-static` includes `BoringSSL` where grpc can use it to establish secure connection. On the other hand, user can also set up `Openssl` locally and use `netty-tcnative` to dynamically link to the local `Openssl` binary to establish secured connection. Is that what's going on?

I think we might need to add doc on how to properly tls on jetcd.  This link is helpful https://github.com/grpc/grpc-java/blob/master/SECURITY.md

lburgazzoli: Correct, I've added netty-tcnative-boringssl-static to test classpath so tests (when implemented) will use the same ssl implementation regardless of the OS

fanminshi:
lgtm

fanminshi:
could you change commit title to
`client: make it easy to configure ssl`? thanks.

lburgazzoli:
done

fanminshi:
Thanks for the pr.

Issue #172, opened 2017-07-14
client: support balancer in jetcd


fanminshi:
etcd has [balancer](https://github.com/coreos/etcd/blob/master/clientv3/balancer.go) manage etcd client connections. Jetcd should have similar thing.

fanminshi:
close via #173

PR #173, opened 2017-07-15
client: support balancer in jetcd #172

lburgazzoli:


src/main/java/com/coreos/jetcd/ClientBuilder.java
```diff
@@ -115,14 +117,36 @@ public ClientBuilder setNameResolverFactory(NameResolver.Factory nameResolverFac
...
   public NameResolver.Factory getNameResolverFactory() {
     return nameResolverFactory;
   }
 
+  /**
+   * config LoadBalancer factory.
+   *
+   * @param loadBalancerFactory etcd LoadBalancer.Factory
+   * @return this builder
+   * @throws NullPointerException if password is null
```
fanminshi: what's `password` in context of `LoadBalancer`?

src/main/java/com/coreos/jetcd/internal/impl/ClientConnectionManager.java
```diff
@@ -63,28 +63,37 @@
...
+            NameResolver.Factory resolverFactory = builder.getNameResolverFactory();
+            if (resolverFactory == null) {
+              resolverFactory = SimpleNameResolverFactory.forEndpoints(builder.getEndpoints());
+            }
+
+            channelBuilder = defaultChannelBuilder();
+            channelBuilder.nameResolverFactory(resolverFactory);
+
+            if (builder.getLoadBalancerFactory() != null) {
+              channelBuilder.loadBalancerFactory(builder.getLoadBalancerFactory());
```
fanminshi: if`LoadBalancerFactory` is not given, what's the default grpc behavior on choosing endpoints?

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.UUID;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+
+/**
+ * KV service test cases.
+ */
+public class LoadBalancerTest {
+  private static final ByteSequence SAMPLE_KEY = ByteSequence.fromString(UUID.randomUUID().toString());
```
fanminshi: you can use `TestUtil.randomString()` to generate random string.

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
```diff
@@ -0,0 +1,67 @@
...
+  private static final ByteSequence SAMPLE_VALUE = ByteSequence.fromString(UUID.randomUUID().toString());
```
fanminshi: use `TestUtil.randomString()` to generate random string

src/main/java/com/coreos/jetcd/internal/impl/ClientConnectionManager.java
lburgazzoli: It uses PickFirstBalancerFactory as default LoadBalancer

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
```diff
@@ -0,0 +1,63 @@
+package com.coreos.jetcd.internal.impl;
+
+import com.coreos.jetcd.ClientBuilder;
+import com.coreos.jetcd.KV;
+import com.coreos.jetcd.kv.PutResponse;
+import io.grpc.PickFirstBalancerFactory;
+import io.grpc.util.RoundRobinLoadBalancerFactory;
+import org.testng.Assert;
+import org.testng.annotations.Test;
```
fanminshi: Do you have a preference on which test frame work to use? I use Junit, assertJ, and mockito. I hope we can agree on one set of test framework so that it will make code review easier. 

e.g https://github.com/coreos/jetcd/blob/master/src/test/java/com/coreos/jetcd/internal/impl/WatchTest.java

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+
+/**
+ * KV service test cases.
+ */
+public class LoadBalancerTest {
+
+  @Test
+  public void testPickFirstBalancerFactory() throws Exception {
+    KV client = ClientBuilder.newBuilder()
```
fanminshi: I prefer to have following testing structures.
1. create client.
2. get *client.
3. do something.
4. close client.

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
```diff
@@ -0,0 +1,63 @@
...
+    long lastMemberId = 0;
+
+    for (int i = 0; i < TestConstants.endpoints.length * 2; i++) {
+      response = client.put(TestUtil.randomByteSequence(), TestUtil.randomByteSequence()).get();
+
+      if (i == 0) {
+        lastMemberId = response.getHeader().getMemberId();
+      }
+
+      Assert.assertEquals(lastMemberId, response.getHeader().getMemberId());
```
fanminshi: I imported [assertJ](http://joel-costigliola.github.io/assertj/) to jetcd for previous testing. I like the fluent style and it is very good for testing exception handling.
e.g
`assertThat(getResponse.getHeader().getMemberId()).isEqualTo(lastMemberId);`

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
```diff
@@ -0,0 +1,63 @@
...
+    KV client = ClientBuilder.newBuilder()
+        .setEndpoints(TestConstants.endpoints)
+        .setLoadBalancerFactory(PickFirstBalancerFactory.getInstance())
+        .build()
+        .getKVClient();
+
+    PutResponse response;
+    long lastMemberId = 0;
+
+    for (int i = 0; i < TestConstants.endpoints.length * 2; i++) {
```
fanminshi: is the reason behind looping twice to the balancer more throughly?

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
```diff
@@ -0,0 +1,63 @@
...
+      response = client.put(TestUtil.randomByteSequence(), TestUtil.randomByteSequence()).get();
+
+      if (i > 0 && lastMemberId != response.getHeader().getMemberId()) {
+        differences++;
+      }
+
+      lastMemberId = response.getHeader().getMemberId();
+    }
+
+    Assert.assertNotEquals(0, differences);
```
fanminshi: use assertJ `assertThat(differences).isNotEqualTo(0);`?

fanminshi:
looks good in general. I just have some question about testing structure.

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
lburgazzoli: No I don't. 

I'm used to JUnit because is the most used one for the project I'm involved but I do not care too much.

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
lburgazzoli: No, is a leftover

lburgazzoli:
Fixed

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
fanminshi: let's use Junit then.

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.coreos.jetcd.Client;
+import com.coreos.jetcd.ClientBuilder;
+import com.coreos.jetcd.KV;
+import com.coreos.jetcd.kv.PutResponse;
+import io.grpc.PickFirstBalancerFactory;
+import io.grpc.util.RoundRobinLoadBalancerFactory;
+import org.testng.annotations.Test;
```
fanminshi: replace the import with `import org.junit.Test;`

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * KV service test cases.
+ */
+public class LoadBalancerTest {
+
```
fanminshi: let's add a timeout rule just in case the test hangs.

```java
  @Rule
  public Timeout timeout = Timeout.seconds(10);
```

fanminshi:
small nits. then lgtm. Thanks!

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
lburgazzoli: So we go with JUnit in future tests ?

fanminshi: yep. I would like to just use one framework.

lburgazzoli:
Fixed

src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
fanminshi: let's have a different Pr to clean up old tests from testNg to junit if we have time.

fanminshi:
lgtm. merge after green.

PR #175, opened 2017-07-15
op: add client endpoint format check

joshua-jin:


src/main/java/com/coreos/jetcd/ClientBuilder.java
```diff
@@ -14,13 +14,17 @@
...
+import java.net.URL;
+import java.util.Arrays;
 import java.util.List;
 
 /**
  * ClientBuilder knows how to create an Client instance.
  */
 public class ClientBuilder implements Cloneable {
-
+  private static final List<String> SCHEMES = Arrays.asList("http", "https", "unix", "unixs");
```
lburgazzoli: Is unix supported by default ?  As far as I can remember the SimpleNameResolver explicit checks for http/https. In addition the endpoint format depends on the NameResolver which is being used so maybe the check should be moved on the *NameResolver* class.

fanminshi: etcd go client supports unix and unixs mainly for testing purpose i think. Probably no need to support that for java client for now. 

cc/ @xiang90

fanminshi: @lburgazzoli for now, I think we should just accept `http/https` and get this pr merged. we can refactor the check into `NameResolver ` in a different pr after https://github.com/coreos/jetcd/issues/176 is discussed. let me know what you think.

lburgazzoli: Fine with me

fanminshi: @joshua-jin please just keep schemes to `["http", "https"]`. then lgtm.

lburgazzoli:
Can you rebase and squash ?

joshua-jin:
@lburgazzoli may be I do some mistake in rebase process. I will close this pr and create a new one.

PR #179, opened 2017-07-18
api: better response wrapper

### LINKED ISSUE #167: Discuss client message wrapper around protobuf obj
Currently, there is a client side response message for every given protobuf msg.
when client receives a response, it copies all the content from the protobuf response to construct the client side response and returns it to the user; the duplication might be expensive.

I would like to discuss a better approach to return client side response.

I took a look `PutResponse` from
https://github.com/lburgazzoli/etcd4j-v3/blob/master/src/main/java/com/github/lburgazzoli/etcd/v3/model/PutResponse.java

It seems like a good approach where PutResponse embeds api.PutResponse. e.g retrieve header can be directly from the api response.

cc/ @lburgazzoli @xiang90 @heyitsanthony

fanminshi:
fixes #167


src/main/java/com/coreos/jetcd/cluster/MemberAddResponse.java
```diff
@@ -34,7 +30,17 @@ public Member getMember() {
...
+    if (members == null) {
+      synchronized (membersRef) {
+        members = membersRef.get();
+        if (members == null) {
+          members = Util.toMembers(getResponse().getMembersList());
+          membersRef.lazySet(members);
+        }
+      }
+    }
+
```
lburgazzoli: Do we need to make this thread safe ? 

I don't think so but if required a simple synchronized method should be enough (we can save some bits as AtomicReference won't be required)

fanminshi: The atomicref used here is to cache the converted result from api.MemberList to Cluster.MemberList. so that subsequent calling of getMembers return the cached value. can you elaborate why `AtomicReference won't be required`?

fanminshi: a simple `synchronized` should be good and make the code cleaner.

lburgazzoli: Yes this is what I meant, it could be as simple as:

```java 
public synchronized List<Member> getMembers() {
    if (this.members == null) {
        this.memebrs = Util.toMembers(getResponse().getMembersList());
    }

    return this.members;
}
```

fanminshi: @lburgazzoli I want cached `Util.toMembers(getResponse().getMembersList())` using `AtomicReference`. 

```java
  public synchronized List<Member> getMembers() {
    List<Member> members = membersRef.get();
    if (members == null) {
      members = Util.toMembers(getResponse().getMembersList());
      membersRef.lazySet(members);
    }

    return members;
  }
```

What do you think about that approach?

lburgazzoli: Does not give any effective advantage as the method is synchronized and it introduces an AtomicReference object which is allocated even not needed.

fanminshi: @lburgazzoli I see what you mean. Having an instance var `member` protected under `synchronized` has same effect has using  `AtomicReference` approach.

fanminshi: will change.

fanminshi:
all fixed. PTAL cc\ @lburgazzoli

lburgazzoli:
lgtm

Issue #176, opened 2017-07-15
[discuss] remove name resolver factory from ClientBuilder


lburgazzoli:
As today we have the possibility to set the name resolver factory on the client builder but it may be better to determine the resolver to use based on the scheme used for the endpoints so i.e.

```java
    // Use the simple resolver
    ClientBuilder.newBuilder().setEndpoints("http://node1:2379", ...)

    // Use the dns+srv resolver
    ClientBuilder.newBuilder().setEndpoints("dns+srv://_etcd-server._tcp.")
```

They can also be mixed and if an advanced user needs to apply advanced configuration to the underlying channel it can provide its own ManagedChannelBuilder as today.

fanminshi:
@lburgazzoli I need sometime to think about this. 
Eventually, I want to add 
`void SetEndpoints(List<String> endpoints)`  function to `client` where this function
allows user to updates endpoints dynamically.  
https://github.com/coreos/etcd/blob/master/clientv3/client.go#L119-L122

if `newBuilder().setEndpoints` only accepts `http/https` then `client.SetEndpoints` only update endpoints for scheme `http/https`. If we introduce more sheme such as `dns+srv` in `newBuilder().setEndpoints(`. Then how what's the expectation of ``client.SetEndpoints` on endpoint with scheme `dns+srv`.

lburgazzoli:
It would do the same: it goes down to the resolver and it is up to the resolver to update the right underlying resolver.

```java
for (MyResolver resolver : resolvers) {
    // each resolver updates its endpoint list
    // only if it can handle the given scheme
    resolver.update(endpoints);
}
```

But I'm not a big fan to have a mixed way to set options as in general it cause confusion, maybe we can have something like:

```java
ImmutableEndpointList endpoints = EndpointsList.immutable(...)
Builder.setEndpoints(endpoints)
```
And:

```java
MutableEndpointList endpoints = EndpointsList.mutable(...)
Builder.setEndpoints(endpoints)

endpoints.update(...)
```

fanminshi:
@lburgazzoli i agree with on this. I don't like mix options too.  the idea of setting resolver based on the scheme  sounds good to me.  cc/ @xiang90

lburgazzoli:
If you guys approve I can work on it

fanminshi:
[...]

PR #181, opened 2017-07-21
resolver: remove name resolver factory from ClientBuilder #176

lburgazzoli:


src/main/java/com/coreos/jetcd/resolver/URIResolvers.java
```diff
@@ -0,0 +1,116 @@
...
+      return Integer.MIN_VALUE;
+    }
+
+    @Override
+    public boolean supports(URI uri) {
+      return "http".equals(uri.getScheme()) || "https".equals(uri.getScheme());
+    }
+
+    @Override
+    public List<EquivalentAddressGroup> resolve(URI uri) {
```
fanminshi: should we also add a check on `uri` to see if it is supported in `resolve`? If not supported then don't add it the the group and cache. Right now it seems to me that the caller can by pass in whatever `uri` that it wants if it forgets to check the `uri` against `supports`.

fanminshi:
looks good in general.

src/main/java/com/coreos/jetcd/resolver/URIResolvers.java
lburgazzoli: The usage pattern would not require this double check but I'll add it for extra safety.

lburgazzoli:
Done

src/main/java/com/coreos/jetcd/resolver/URIResolvers.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @Override
+    public int priority() {
+      return Integer.MIN_VALUE;
+    }
+
+    @Override
+    public boolean supports(URI uri) {
+      return "http".equals(uri.getScheme()) || "https".equals(uri.getScheme());
+    }
```
fanminshi: we can use checking code in `isValidEndpointFormat` for Direct resolver's `supports`.

e.g

```java
private static final List<String> DRIECT_SCHEMES = Arrays.asList("http", "https");

public boolean supports(URI uri) {
    if (SCHEMES.stream().noneMatch(u.getProtocol()::equals)) {		
      return false;		
    }
    
    // endpoint must contain a port.		
    if (u.getPort() == -1) {		
      return false;		
    }
    
    // endpoint must not contain a path.		
    return u.getPath().isEmpty();	
}
```

src/test/java/com/coreos/jetcd/resolver/URIResolversTest.java
```diff
@@ -0,0 +1,35 @@
...
+    URIResolvers.Direct discovery = new URIResolvers.Direct();
+
+    for (String scheme : DIRECT_SCHEMES) {
+      URI uri = URI.create(scheme + "://127.0.0.1:2379");
+
+      assertThat(discovery.supports(uri)).isTrue();
+      assertThat(discovery.resolve(uri).size()).isGreaterThan(0);
+    }
+  }
+
```
fanminshi: Can we add tests for invalid urls for Direct Resolvers.

transform those deleted tests to here.

```java
-  @Test(dataProvider = "invalidEndpoints", expectedExceptions = IllegalArgumentException.class)		
 -  public void testBuild_WithInValidFormatEndpoints(String endpoint) {		
 -    builder.setEndpoints(endpoint);		
 -  }		
 -		
 -  @DataProvider		
 -  private Object[][] invalidEndpoints() {		
 -    return new Object[][] {		
 -        {"://127.0.0.1:2379"},		
 -        {"mailto://127.0.0.1:2379"},		
 -        {"http://127.0.0.1"},		
 -        {"http://127.0.0.1:2379/path"}		
 -    };		
 -  
```

fanminshi:
few concerns then lgtm.

lburgazzoli:
Fixed now

PR #189, opened 2017-08-01
 watch: throw etcd exception about watcher being closed and make watch and watcher close idempotent

### LINKED ISSUE #180: watcher.close() throw the EtcdException(InterruptedException)
Jetcd will throw an `EtcdException` if you close an on going listened watcher.
The `InterruptedException` exception caused by `this.eventsQueue.take();` when you interrupted the thread.
For `this.eventsQueue.take();` , it is an exception if the thread is interrupted. But for Jetcd the expected behavior for the `watcher.close()` method is to get the thread interrupted. So why does Jetcd raise an `EtcdException` exception?

`Caused by: java.lang.InterruptedException
	at java.util.concurrent.FutureTask.awaitDone(FutureTask.java:404)
	at java.util.concurrent.FutureTask.get(FutureTask.java:191)
	at com.coreos.jetcd.internal.impl.WatchImpl$WatcherImpl.listen(WatchImpl.java:449)
	... 4 more`

fanminshi:
fixes #180

### DIFF
--- a/src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
+++ b/src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
@@ -87,7 +87,9 @@ class WatchImpl implements Watch {
 
   @Override
   public synchronized void close() {
-    Preconditions.checkState(!isClosed(), "Watch client has been closed");
+    if (isClosed()) {
+      return;
+    }
 
     this.setClosed();
     this.notifyWatchers(EtcdExceptionFactory.newEtcdException("Watch client has been closed"));
@@ -419,7 +421,9 @@ class WatchImpl implements Watch {
     @Override
     public void close() {
       synchronized (this.closedLock) {
-        Preconditions.checkState(!isClosed(), "Watcher has been closed");
+        if (isClosed()) {
+          return;
+        }
 
         this.setClosed();
       }
@@ -435,6 +439,12 @@ class WatchImpl implements Watch {
       try {
         return this.createWatchResponseFuture().get();
       } catch (ExecutionException e) {
+        synchronized (this.closedLock) {
+          if (isClosed()) {
+            // returns EtcdException indicates watcher has been closed.
+            throw EtcdExceptionFactory.newEtcdException("Watcher has been closed");
+          }
+        }
         Throwable t = e.getCause();
         if (t instanceof CompactedException) {
           throw (CompactedException) t;


fanminshi:
cc/ @jalenyang.  I changed the code to return a empty response if watch is closed.

jalenyang:
@fanminshi I think the change should be something like this:
`@Override
        public synchronized com.coreos.jetcd.watch.WatchResponse listen() {
            Preconditions.checkState(!isClosed(), "Watcher has been closed");

            try {
                return this.createWatchResponseFuture().get();
            } catch (ExecutionException e) {
                Throwable t = e.getCause();
                if (t instanceof CompactedException) {
                    throw (CompactedException) t;
                }
                if (t instanceof EtcdException) {
                    throw (EtcdException) t;
                }
                throw EtcdExceptionFactory.newEtcdException(t);
            } catch (InterruptedException e) {
                synchronized (this.closedLock) {
                    if (isClosed()) {
                        // watcher closed while listening.
                        // return a default response.
                        return com.coreos.jetcd.watch.WatchResponse.getDefaultInstance();
                    }
                }
            } catch (Exception e) {
                throw EtcdExceptionFactory.newEtcdException(e);
            }
        }`

fanminshi:
@jalenyang your suggestion won't solve the issue you had. Closing watcher interrupts the underlying executor which is then wrapped in `ExecutionException `.  That's why my code check under `ExecutionException`. Your suggestion is something else. Suppose that your calling thread is waiting on ` this.createWatchResponseFuture().get();` then that thread is interrupt by your executor that contains the thread. Then ` } catch (InterruptedException e) {` will be catched.

fanminshi:
Take a look of my test.
```java
  public void testWatcherListenOnWatcherClose() {
     Watcher watcher = watchClient.watch(KEY);
     executor.execute(() -> {
       try {
         Thread.sleep(20);
         watcher.close(); // closes watcher's internal executor.
       } catch (InterruptedException e) {
         // do nothing
       }
     });
    // asstertThat take response from watcher.listen() and compare it to com.coreos.jetcd.watch.WatchResponse.getDefaultInstance()
     assertThat(watcher.listen())
         .isEqualTo(com.coreos.jetcd.watch.WatchResponse.getDefaultInstance());
   }
```

jalenyang:
@fanminshi yes, you are right. You caught the `ExecutionException ` and return the default instance before
`InterruptedException`.
Thanks!

fanminshi:
@jalenyang a second thought. I don't returning an empty response make sense. But I should return a EtcdException that indicate that watcher is being closed instead of being interrupted. Anyways, you should have a mechanism of handling watcher exceptions.

```java
    ...
    try {
      WatchResponse response = watcher.listen();
      ...
    } catch (CompactedException e) {
      long compactedRev = e.getCompactedRevision();
      // you can create a new watcher starting at rev > compactedRev.
    } catch (EtcdException e) {
      // log it?
    } finally {
      watcher.close();
    }
```

xiang90:
lgtm

PR #191, opened 2017-08-03
exception: enhance etcd exception

### LINKED ISSUE #185: discuss: provide a more detail jetcd exception message.
the current EtcdException only contains `String message` and  `Throwable cause`
```java
public class EtcdException extends RuntimeException {
  EtcdException(String message, Throwable cause) {
    super(message, cause);
  }
}
```
I noticed that the etcd4j project has a more detail EtcdExcpetion which includes errorCode as follow
```java
public class EtcdException extends Exception {
...
  public final String etcdCause;
  public final int errorCode;
  public final Long index;
  public final String etcdMessage;

  /**
   * Constructor
   *
   * @param errorCode the etcd error code
   * @param cause     the exception cause
   * @param message   the exception message
   * @param index     the data index
   */
  protected EtcdException(
    @JsonProperty("errorCode") Integer errorCode,
    @JsonProperty("cause") String cause,
    @JsonProperty("message") String message,
    @JsonProperty("index") Long index) {
    this.errorCode = errorCode;
    this.etcdCause = cause;
    this.etcdMessage = message;
    this.index = index;
  }
...
```
https://github.com/jurmous/etcd4j/blob/master/src/main/java/mousio/etcd4j/responses/EtcdException.java

In addition, I noticed that cloud spanner api have similar pattern.
```java
public class SpannerException extends BaseGrpcServiceException {
...
  private final ErrorCode code;

  /** Private constructor. Use {@link SpannerExceptionFactory} to create instances. */
  SpannerException(
      DoNotConstructDirectly token,
      ErrorCode code,
      boolean retryable,
      @Nullable String message,
      @Nullable Throwable cause) {
    super(message, cause, code.getCode(), retryable);
    if (token != DoNotConstructDirectly.ALLOWED) {
      throw new AssertionError("Do not construct directly: use SpannerExceptionFactory");
    }
    this.code = Preconditions.checkNotNull(code);
  }

  /** Returns the error code associated with this exception. */
  public ErrorCode getErrorCode() {
    return code;
  }
...
```
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerException.java

I wonder if jetcd should follow the above way of defining exception?

fanminshi:
fixes #185


fanminshi:
Work in progress.
This pr depends on merging of https://github.com/coreos/jetcd/pull/183.

fanminshi:
Please mainly review for the new `EtcdException`, `ErrorCode`, and` Constants` classes and on whether or not they are being used appropriately.  cc/ @lburgazzoli.

lburgazzoli:
I didn't go down to each single change but it would be nice to know what is the rationale as I see a few critical points:

- Constants hold a huge number of pre-instantiated exceptions that just waste memory and as exception are 'exceptional' we should create them only when needed.
- Constants has a number of hard-coded messages that may result in maintenance efforts as they have to be kept in sync with etcd and IMHO we should rely on error codes as much as possible and leave the message to be provided by etcd as much as possible.
- EtcdExceptionFactory.toEtcdException may breaks the throwable chain by removing relevant information till an exception of type EtcdException is found which is bad.
- We should carefully wrap exceptions as it make it harder to properly catch exceptions, i.e. if all the exceptions are of type EtcdException, how can I catch an InterruptedException ? Rely on getCause() is quite ugly

fanminshi:
My rationale for having the constant so that user can compare exception being thrown. e.g
```java
    Watcher watcher = watchClient.watch(KEY);

    try {
      watcher.listen();
    } catch (EtcdException e) {
      if (e == Constants.CLOSED_WATCHER) {
        // ignore
      }
      if (e == Constants.FUTURE_REV) {
        // log it?
      }
    }
```

However, I am unsure if user care about comparing `EtcdException`. I would like your opinion on this.

`EtcdExceptionFactory.toEtcdException may breaks the throwable chain by removing relevant information till an exception of type EtcdException is found which is bad.`
If we stick with error constants, then this is an issue. However, if it is removed, then this shouldn't be a problem.

`We should carefully wrap exceptions as it make it harder to properly catch exceptions, i.e. if all the exceptions are of type EtcdException, how can I catch an InterruptedException ? Rely on getCause() is quite ugly`

I followed spanner api on how to handle exception. it seems me that all exceptions thrown are type of SpannerException including interrupt. 
```java
  public static SpannerException propagateInterrupt(InterruptedException e) {
    Thread.currentThread().interrupt();
    return SpannerExceptionFactory.newSpannerException(ErrorCode.CANCELLED, "Interrupted", e);
  }
```
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerExceptionFactory.java#L46-L49 
If all the exception is one type, I wonder whether it is easier to reason about the error that having multiple exceptions. 

for jetcd, the some api is blocking. We can throw checked Interrupted Exception for all those. 

e.g

```java
interface Watcher {

  /**
   * closes this watcher and all its resources.
   **/
  void close();

  /**
   * Retrieves next watch key, waiting if there are none.
   *
   * @throws com.coreos.jetcd.exception.EtcdException when listen encounters connection error,
   *        watcher closed, watch client closed, etcd server error, and any internal client error.
   * @throws com.coreos.jetcd.exception.CompactedException when watch a key at a revision that has
   *        been compacted.
   * @throws InterruptedException if listening thread is interrupted.
   * 
   */
  WatchResponse listen() throws InterruptedException;;
}
```

Also then we can go into unchecked vs checked exception. The book clean code suggests to use unchecked exception for all exception if I recall correctly also I have learned that google internally only throw unchecked exception as we can see that pattern from spanner apis.

In conclusion, I decided to go with one exception type which is EtcdException also unchecked to follow spanner and google style. For Error Constants, I added it so that the user can compare EtcdExceptions.

@lburgazzoli Let me know what you think on how jetcd do error handling.

lburgazzoli:
Comparing object with == is dangerous as it compare by reference so it is true only if the two object are the same, in other words it is like comparing pointers so any use of 

```java 
throw EtcdExceptionFactory.newEtcdException(e)
```
would invalidate the assumption.

User should rely on code and exception type so we should have a generic EtcdException and maybe some specialized exceptions easier to catch i.e.

```java
    try {
      watcher.listen();
    } catch (SpecializedEtcdException e) {
        // do something
    }
```

This would be the same as

```java
    try {
      watcher.listen();
    } catch (EtcdException e) {
        if (e.getCode() == SPECIALIZED_EXCEPTION) {
            //do something
        } else {
            throw e;
        }
    } 
```
but with less noises.

Btw, I'm ok with unchecked exception.

fanminshi:
` if a user callback throws InterruptedException this should not be wrapped as EtcdException asn it does not comes from the etcd stack.`
that makes sense. then I'll change blocking call to throw interrupt exception if it is not from etcd stack.

fanminshi:
@lburgazzoli Get rid of the error constants. PTAL.

src/main/java/com/coreos/jetcd/exception/ClosedSnapshotException.java
```diff
@@ -0,0 +1,19 @@
+package com.coreos.jetcd.exception;
+
+public class ClosedSnapshotException extends EtcdException {
+
+  private static final ClosedSnapshotException DEFAULT_INSTANCE;
+
+  static {
+    DEFAULT_INSTANCE = new ClosedSnapshotException();
```
lburgazzoli: Don't do that  as the stack trace of the exception will be initialized with the location of the first call, an subsequent throw will report the same stack regardless of the location in which you invoke ``throw XyzException.getDefaultInstance()``

src/main/java/com/coreos/jetcd/exception/ErrorCode.java
```diff
@@ -0,0 +1,66 @@
...
+    } catch (IllegalArgumentException e) {
+      return defaultValue;
+    }
+  }
+
+  /**
+   * Returns the error code corresponding to a gRPC status, or {@code UNKNOWN} if not recognized.
+   */
+  static ErrorCode fromGrpcStatus(Status status) {
+    ErrorCode code = errorByRpcCode.get(status.getCode().value());
```
lburgazzoli: Not sure if we need ``errorByRpcCode`` as we can loop through the values, similar to what is done in the static initialization block.

src/main/java/com/coreos/jetcd/internal/impl/LeaseImpl.java
```diff
@@ -313,13 +322,12 @@ private LeaseKeepAliveRequest newKeepAliveRequest(long leaseId) {
   }
 
 
-  private class KeepAliveListenerImpl implements KeepAliveListener {
+  private static class KeepAliveListenerImpl implements KeepAliveListener {
```
lburgazzoli: Any reason to make it static ?

src/main/java/com/coreos/jetcd/exception/ClosedSnapshotException.java
fanminshi: I guess I will just have to create a new instance of the `ClosedSnapshotException` every time then.

src/main/java/com/coreos/jetcd/exception/ErrorCode.java
fanminshi: probably no need. Will remove.

src/main/java/com/coreos/jetcd/internal/impl/LeaseImpl.java
fanminshi: I was playing static bug checker. It complains about the nested class not being static. It seems to be the best practice in java is to have inner static class.
https://stackoverflow.com/questions/19825956/java-inner-class-access-and-best-practices

lburgazzoli: Not always, sometime, especially with implementations, have a non static inner class is better as in can access instance variables.

src/main/java/com/coreos/jetcd/exception/ErrorCode.java
fanminshi: actually, will `errorByRpcCode` be faster since it is a constant operation to look up error?  I guess it is an optimization.

src/main/java/com/coreos/jetcd/internal/impl/LeaseImpl.java
fanminshi: @lburgazzoli I see. In my case, I realize that I don't need to access the inner class variable. I guess making the inner class static make sense for now.

Issue #194, opened 2017-08-08
issue: maven install task doesn't copy services directory into META-INF


joshua-jin:
After I use `mvn install` to build target jar, I found `service` directory and file `com.coreos.jetcd.resolver.URIResolver` which under this directory are not copied into `META-INF`. It will make grpc calls failure, because of uri resolve issue.

PR #196, opened 2017-08-08
issue: maven install task doesn't copy services directory into META-INF #194

lburgazzoli:


fanminshi:
@lburgazzoli which one is the prefer way to add `META-INF`? this or https://github.com/coreos/jetcd/pull/199/?

PR #199, opened 2017-08-08
Bundle: add META-INF/services content

### LINKED ISSUE #194: issue: maven install task doesn't copy services directory into META-INF

paoloantinori:
fixes #194

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,6 +122,14 @@
     </dependencies>
 
     <build>
+        <resources>
+          <resource>
+            <directory>target/classes</directory>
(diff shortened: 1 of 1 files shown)


lburgazzoli:
lgtm

lburgazzoli:
thank you @paoloantinori

Issue #83, opened 2017-04-23
static check using ErrorProne


heyitsanthony:
https://github.com/google/error-prone

fanminshi:
closed via https://github.com/coreos/jetcd/pull/203

PR #203, opened 2017-08-09
build: static check using ErrorProne #83

lburgazzoli:
error-prone added to build reactor
fix error-prone reports


fanminshi:
lgtm

PR #211, opened 2017-08-11
pom: add shade plugin to generate uber jar

fanminshi:


fanminshi:
merge this with https://github.com/coreos/jetcd/pull/209

PR #214, opened 2017-08-17
project: layout refactoring

lburgazzoli:
This refactoring includes the following directories:

- jetcd-core with the main source tree
- jetcd-examples with a simple watch example
- jetcd-all which build an uber-jar with packages relocation
  (OSGi support is not yet complete)
- etc which contains additional stuffs like checkstyle and
  scripts to run etcd cluster among wich I provided one to
  run a cluster in docker


fanminshi:
lgtm once ci passes.

PR #220, opened 2017-08-22
example: add a simple etcdctl example using jetcd client

fanminshi:
add jetcd-simple-ctl example to illustrate how to build a simple ctl of GET, PUT, and WATCH using jetcd client.

### DIFF
--- a/jetcd-examples/jetcd-simple-ctl/README.MD
+++ b/jetcd-examples/jetcd-simple-ctl/README.MD
@@ -1,8 +1,8 @@
 ### jetcd-watch-example
 
-Start an etcd cluster .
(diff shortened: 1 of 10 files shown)


fanminshi:
PTAL cc/ @lburgazzoli

lburgazzoli:
not sure jetcd-examples/jetcd-simple-ctl/mvnw  is needed

fanminshi:
@lburgazzoli with mvnw, it makes the README examples easier to follow or else i'll assume the the user to have `mvn` installed.

lburgazzoli:
So you need mvnw.bat too ;)

fanminshi:
@lburgazzoli true. what's do you think it is best way to run the example then?

fanminshi:
@lburgazzoli added support for windows

PR #227, opened 2017-08-24
pom: remove parent pom dependency for examples

fanminshi:
examples should have independent pom configuration that doesn't depend on jetcd-parent


lburgazzoli:
not sure about that, it is usually better to have a parent to set common properties, plugin, etc.

fanminshi:
@lburgazzoli the issue I am seeing is that I expect that user can just copy and paste the pom.xml from the examples and expect it to work. With example pom.xml depends on parent-pom.xml, there is an extra dependency that user needs to reason about. 

For instance, the blog I am writing describe how to build the jetcd-simple-ctl from scratch. If its pom.xml depends on jetcd-parent pom, user might not understand what's going on.

lburgazzoli:
not sure. i.e. the example contains also the shade plugin which won't be used 99% of the time as most of the time people will embed jetcd in theirs app.

PR #228, opened 2017-08-25
jetcdctl: improve error handling

heyitsanthony:
* exit code 1 on command failures
* don't throw an out of bounds exception if getting a key that does not exist; print nothing like etcdctl


fanminshi:
lgtm

PR #241, opened 2017-09-29
build: update plugin and dependencies

lburgazzoli:


pom.xml
```diff
@@ -229,6 +234,36 @@
         <pluginManagement>
             <plugins>
                 <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
```
fanminshi: does maven include those clean, deploy, install by default?

lburgazzoli: yes. the reason I explicit set the version is that the default version is set in the root pom and may vary according to the maven version so it is a good idea to fix it in order to have the same behavior regardless of the maven version used.

fanminshi: sounds good!

fanminshi:
lgtm

PR #244, opened 2017-09-29
client: add AutoCloseable

ScienJus:
for https://github.com/coreos/jetcd/issues/239

I think `Client` , `CloseableClient` and `Watcher` should extends `AutoCloseable`.

but i can't make sure it's necessary for `KeepAlive` and `KeepAliveListener`, because i don't know how to use them.


jetcd-core/src/main/java/com/coreos/jetcd/Watch.java
```diff
@@ -48,11 +48,12 @@
...
    * Interface of Watcher.
    */
-  interface Watcher {
+  interface Watcher extends AutoCloseable {
 
     /**
      * closes this watcher and all its resources.
      **/
+    @Override
     void close();
```
lburgazzoli: AutoCloseable::close() throws an exception, I do not think we should change the signature

jetcd-core/src/main/java/com/coreos/jetcd/internal/impl/CloseableClient.java
```diff
@@ -16,10 +16,12 @@
...
 
-public interface CloseableClient {
+public interface CloseableClient extends AutoCloseable {
+
   /**
    * close the client and release its resources.
    */
+  @Override
   default void close() {
     // noop
```
lburgazzoli: Same as before

jetcd-core/src/main/java/com/coreos/jetcd/Watch.java
ScienJus: I think the point is whether to throw out exception on `Watch#close`, so far it’s prohibited.

We uses `AutoCloseable` just for `try-with-resource`, it should not be a constraint.

lburgazzoli: so lgtm :)

jetcd-core/src/test/java/com/coreos/jetcd/internal/impl/WatchTest.java
```diff
@@ -26,14 +24,17 @@
...
 import com.coreos.jetcd.watch.WatchResponse;
-import java.util.Arrays;
-import java.util.concurrent.ExecutionException;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.Timeout;
 
+import java.util.Arrays;
```
fanminshi: why move imports around?

jetcd-core/src/test/java/com/coreos/jetcd/internal/impl/LoadBalancerTest.java
```diff
@@ -26,6 +24,8 @@
 import org.junit.Test;
 import org.junit.rules.Timeout;
 
+import static org.assertj.core.api.Assertions.assertThat;
```
fanminshi: prefer not move imports order

jetcd-core/src/main/java/com/coreos/jetcd/Watch.java
```diff
@@ -48,11 +48,12 @@
   /**
    * Interface of Watcher.
    */
-  interface Watcher {
+  interface Watcher extends AutoCloseable {
```
fanminshi: why `extends` not `implement` the  `AutoCloseable`?

if we `@Override` the `close()` from `AutoCloseable`, how come our `close()` doesn't `throw Exception`?

ScienJus: 1. `interface` only can 'extends' another interface. not `implement`.

2. The overriding method can throw those checked exceptions, which have less scope than the exception(s) declared in the overridden method.

fanminshi: now i understand what you were doing. pretty cool java trick.

fanminshi:
lgtm

Issue #162, opened 2017-07-12
Add tests for ssl/tls


lburgazzoli:

fanminshi:
https://github.com/grpc/grpc-java/blob/master/interop-testing/src/test/java/io/grpc/testing/integration/TlsTest.java

there is a example of TlsTest from grpc. Maybe we can get some idea for that.

yangliucheng:
I use the example create a HTTPS client for JETCD,but it did not work very well:
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
java.lang.IllegalArgumentException: File does not contain valid private key: D:\kubernetes\server_key_decode.pem
	at io.netty.handler.ssl.SslContextBuilder.keyManager(SslContextBuilder.java:267)
	at io.netty.handler.ssl.SslContextBuilder.keyManager(SslContextBuilder.java:222)
	at io.netty.handler.ssl.SslContextBuilder.forServer(SslContextBuilder.java:54)
	at com.yangliucheng.com.dilimiter.cache.DiEtcd.<init>(DiEtcd.java:31)
	at com.yangliucheng.com.dilimiter.cache.DiEtcd.create(DiEtcd.java:59)

DamianCummins:
@lburgazzoli 
@fanminshi 

I'm trying to migrate from etcd v2 to v3 using the jetcd library.

However, when I attempt to connect using the SSL certificate (caCert in the code below) I hit the following error:
```
com.coreos.jetcd.exception.EtcdException: Channel closed while performing protocol negotiation
at com.coreos.jetcd.exception.EtcdExceptionFactory.newEtcdException(EtcdExceptionFactory.java:34)
at com.coreos.jetcd.exception.EtcdExceptionFactory.fromStatus(EtcdExceptionFactory.java:82)
at com.coreos.jetcd.exception.EtcdExceptionFactory.toEtcdException(EtcdExceptionFactory.java:78)
at com.coreos.jetcd.exception.EtcdExceptionFactory.toEtcdException(EtcdExceptionFactory.java:73)
at com.coreos.jetcd.internal.impl.ClientConnectionManager.generateToken(ClientConnectionManager.java:236)
at com.coreos.jetcd.internal.impl.ClientConnectionManager.getToken(ClientConnectionManager.java:112)
at com.coreos.jetcd.internal.impl.ClientConnectionManager.access$100(ClientConnectionManager.java:56)
[...]

PR #245, opened 2017-09-29
test(ssl): add tests for ssl/tls #162

lburgazzoli:
fixes https://github.com/coreos/jetcd/issues/162 https://github.com/coreos/jetcd/issues/164


etc/cert/ca-config.json
```diff
@@ -0,0 +1 @@
+{"signing":{"default":{"expiry":"43800h","usages":["signing","key encipherment","server auth","client auth"]}}}
```
fanminshi: is it possible to have cert/* to be under test/resources/cert? since it is part of test?

etc/scripts/generate-self-signed-certificates.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
```
fanminshi: best if it can be under test/resources/folder

jetcd-core/pom.xml
```diff
@@ -181,6 +181,18 @@
 
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <systemPropertyVariables>
+                        <ssl.cert.path>${project.basedir}/../etc/cert</ssl.cert.path>
+                        <ssl.cert.authority>etcd-ssl</ssl.cert.authority>
+                        <ssl.cert.endpoints>https://127.0.0.1:42379</ssl.cert.endpoints>
```
fanminshi: I don't quite understand what this part is doing. could you elaborate?
also I am not sure of the official pom.xml should have configs that'sspecific to ssl tests.

lburgazzoli: They are here to avoid hard coding in tests, you may have your own instance and certs you want to use but that can be improved.

lburgazzoli:
Will rework the pr

lburgazzoli:
@fanminshi anything more I need to change ?

jetcd-core/src/test/resources/ssl/.gitignore
```diff
@@ -0,0 +1 @@
+cfssl/
```
fanminshi: not needed?

pom.xml
```diff
@@ -319,6 +319,13 @@
                             <arg>-Xlint:all</arg>
                             <!-- TODO <arg>-Werror</arg> -->
                         </compilerArgs>
+                        <annotationProcessorPaths>
+                            <annotationProcessorPath>
+                                <groupId>com.google.auto.service</groupId>
```
fanminshi: is autoservice being used anywhere?

jetcd-core/src/test/java/com/coreos/jetcd/internal/impl/SslTest.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Objects;
+import org.junit.Test;
+
+public class SslTest {
+
+  @Test(timeout = 5000)
+  public void testSimpleSllSetup() throws Exception {
+    final ByteSequence key = ByteSequence.fromString(TestUtil.randomString());
+    final ByteSequence val = ByteSequence.fromString(TestUtil.randomString());
+    final String capath = System.getProperty("ssl.cert.capath");
```
fanminshi: are we not using property file anymore?

jetcd-core/src/test/java/com/coreos/jetcd/internal/impl/SslTest.java
```diff
@@ -0,0 +1,63 @@
...
+    final String authority = System.getProperty("ssl.cert.authority", "etcd-ssl");
+    final String endpoints = System.getProperty("ssl.cert.endpoints", "https://127.0.0.1:42379");
```
fanminshi: maybe have a test constant for ssl endpoint?

jetcd-core/src/test/java/com/coreos/jetcd/internal/impl/SslTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    try (InputStream is = Objects.nonNull(capath)
+          ? new FileInputStream(new File(capath))
+          : getClass().getResourceAsStream("/ssl/cert/ca.pem")) {
```
fanminshi: I think we are not using property to get the capth anymore. maybe just `getClass().getResourceAsStream("/ssl/cert/ca.pem")`? will be enough?

.gitignore
```diff
@@ -1,5 +1,6 @@
 .*
 !.gitignore
+!.travis.yml
```
fanminshi: why adding this?

fanminshi:
@lburgazzoli looks good in general. I just have few questions.

.gitignore
lburgazzoli: Because travis.yml should not be ignored :) but it is by default because of .*

jetcd-core/src/test/java/com/coreos/jetcd/internal/impl/SslTest.java
lburgazzoli: We still have properties so you can test against your own secured etcd instance setting properties

jetcd-core/src/test/java/com/coreos/jetcd/internal/impl/SslTest.java
lburgazzoli: I want to be able to test against my own etcd setup with my own ca, so this is still needed

jetcd-core/src/test/resources/ssl/.gitignore
lburgazzoli: It is needed only to generate the certs and it is created by the script

pom.xml
lburgazzoli: I can try to move it on jetcd-core but it does not hurt :)

lburgazzoli:
fixed

jetcd-core/src/test/resources/ssl/.gitignore
fanminshi: I see.

fanminshi:
lgtm

fanminshi:
@lburgazzoli thanks for the fix! appreciated

fanminshi:
merge when green.

Issue #249, opened 2017-10-15
Name resolving seems overcomplicated


gtudan:
The `SmartNameResolver` uses the Java ServiceLoader to lookup implementations of `URIResolver`. I would like to understand why this design decision was made, as it seems a little overengineered. The implementations are right there in the same package. Is there a use case for implementing your own resolver?

I tripped over this one when I tried started implementing a JCA-Resource-Adapter based on this library. Classloading in Application Servers can be quiet messy and the service loader caught me off guard.

If you had a use-case for this in mind then I'm fine with the service loader - but could we default to the build-in ones in case the service-loader fails to resolve them?

lburgazzoli:
The initial idea was to move additional resolvers like dns-srv out of the core package that's why we have the service loader now but this is not the final version as we have also to deal with OSGi so cassloading is definitively an issue.

I'll try to make it simple and flexible in the coming days.

lburgazzoli:
It should be fixed by #250

PR #250, opened 2017-10-16
Name resolving seems overcomplicated #249

lburgazzoli:
- dns-srv resolver is now an optional packages
- the direct resolver is now loaded by default
- service loader is now used by default but the logic to discover resolvers is pluggable


lburgazzoli:
@fanminshi do you ind reviewing it ?

fanminshi:
@lburgazzoli sorry for the delay. I am reviewing it now.

jetcd-core/src/main/java/com/coreos/jetcd/internal/impl/ClientConnectionManager.java
```diff
@@ -159,7 +160,9 @@ synchronized void close() {
         .nameResolverFactory(
```
fanminshi: it seems to be setting `nameResolverFactory` again is not needed. 
`defaultChannelBuilder()` has already set the `nameResolverFactory`.

jetcd-discovery-dns-srv/pom.xml
```diff
@@ -0,0 +1,139 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
fanminshi: I am not sure if main jetcd repo should have dns resolver implementation. the users can write one themself.
I think its better to keep code base simple and small.

jetcd-core/src/main/java/com/coreos/jetcd/resolver/URIResolverLoader.java
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * Creates a default implementation based on {@link ServiceLoader}.
+   *
+   * @return the default implementation.
+   */
+  static URIResolverLoader defaultLoader() {
+    return () -> {
+      List<URIResolver> resolvers = new ArrayList<>();
+
+      ServiceLoader.load(URIResolver.class).forEach(resolvers::add);
```
fanminshi: How does `ServiceLoader` find  `DirectUriResolver `? I notcied that auto service for `DirectUriResolver` is removed.

fanminshi:
@lburgazzoli sorry for the delay. I have reviewed the pr.

jetcd-core/src/main/java/com/coreos/jetcd/internal/impl/ClientConnectionManager.java
lburgazzoli: There is a small difference between the two: 
- `defaultChannelBuilder` uses the endpoints from the builder
- `withNewChannel` provides its own endpoint so we need to override the `nameResolverFactory`

jetcd-core/src/main/java/com/coreos/jetcd/resolver/URIResolverLoader.java
lburgazzoli: The DirectUriResolver is included by default, see https://github.com/lburgazzoli/coreos-jetcd/blob/github-249/jetcd-core/src/main/java/com/coreos/jetcd/resolver/SmartNameResolver.java#L63

jetcd-discovery-dns-srv/pom.xml
lburgazzoli: Well, it was already included before so I do not see the point to remove it and I think we should provide reasonable default implementation for common things like this one.

jetcd-core/src/main/java/com/coreos/jetcd/resolver/DirectUriResolver.java
```diff
@@ -0,0 +1,79 @@
...
+  @Override
+  public boolean supports(URI uri) {
+    if (!SCHEMES.contains(uri.getScheme())) {
+      return false;
+    }
+
+    if (!Strings.isNullOrEmpty(uri.getPath())) {
+      return false;
+    }
+
```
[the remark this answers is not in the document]
lburgazzoli: fixed

jetcd-core/src/main/java/com/coreos/jetcd/resolver/DirectUriResolver.java
```diff
@@ -0,0 +1,79 @@
...
+          "Unsupported URI " + uri
+      );
+    }
+
+    return this.cache.computeIfAbsent(
+        uri,
+        u -> Collections.singletonList(
+            new EquivalentAddressGroup(new InetSocketAddress(
+                uri.getHost(),
+                uri.getPort() != -1 ? uri.getPort() : 2739)
```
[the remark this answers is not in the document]
lburgazzoli: fixed

jetcd-discovery-dns-srv/pom.xml
fanminshi: I see.

fanminshi:
lgtm. thanks @lburgazzoli

PR #259, opened 2017-11-24
GG-489/#246 add lock support

### LINKED ISSUE #246: Support of lock API
Cant it be implemented?

https://github.com/coreos/etcd/blob/master/Documentation/dev-guide/api_concurrency_reference_v3.md#service-lock-etcdserverapiv3lockv3lockpbv3lockproto

Grisu118:
I added a lock client which supports the API of the etcd lock service ([doc](https://github.com/coreos/etcd/blob/master/Documentation/dev-guide/api_concurrency_reference_v3.md#service-lock-etcdserverapiv3lockv3lockpbv3lockproto))

fixes #246

### DIFF
--- a/jetcd-core/src/main/java/com/coreos/jetcd/Lock.java
+++ b/jetcd-core/src/main/java/com/coreos/jetcd/Lock.java
@@ -1,0 +1,52 @@
+/**
+ * Copyright 2017 The jetcd authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.coreos.jetcd;
+
+import com.coreos.jetcd.data.ByteSequence;
+import com.coreos.jetcd.internal.impl.CloseableClient;
+import com.coreos.jetcd.lock.LockResponse;
+import com.coreos.jetcd.lock.UnlockResponse;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Interface of Lock talking to etcd.
+ */
+public interface Lock extends CloseableClient {
+
+  /**
+   * Acquire a lock with the given name.
+   *
+   * @param name
+   *          the identifier for the distributed shared lock to be acquired.
+   * @param leaseId
+   *          the ID of the lease that will be attached to ownership of the
+   *          lock. If the lease expires or is revoked and currently holds the
+   *          lock, the lock is automatically released. Calls to Lock with the
+   *          same lease will be treated as a single acquistion; locking twice
+   *          with the same lease is a no-op.
+   */
+  CompletableFuture<LockResponse> lock(ByteSequence name, long leaseId);
+
+  /**
+   * Release the lock identified by the given key.
+   *
+   * @param lockKey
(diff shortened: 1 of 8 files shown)


jetcd-core/src/main/java/com/coreos/jetcd/Lock.java
```diff
@@ -0,0 +1,52 @@
...
+   * @param name
+   *          the identifier for the distributed shared lock to be acquired.
+   * @param leaseId
+   *          the ID of the lease that will be attached to ownership of the
+   *          lock. If the lease expires or is revoked and currently holds the
+   *          lock, the lock is automatically released. Calls to Lock with the
+   *          same lease will be treated as a single acquistion; locking twice
+   *          with the same lease is a no-op.
+   */
+  CompletableFuture<LockResponse> lock(ByteSequence name, long leaseId);
```
lburgazzoli: Is lease always required ? if not I think it would be nice to have an overloaded method like: 

    CompletableFuture<LockResponse> lock(ByteSequence name)

jetcd-core/src/main/java/com/coreos/jetcd/data/ByteSequence.java
```diff
@@ -67,7 +67,7 @@ public boolean equals(Object obj) {
     }
   }
 
-  protected ByteString getByteString() {
+  public ByteString getByteString() {
```
lburgazzoli: This method should not be made public as you can then change the content of the `byteString` which invalidates the hash code calculation done in the constructor.

jetcd-core/src/main/java/com/coreos/jetcd/Lock.java
Grisu118: I think it's not required, but in my opinion such a method makes no sense. Because you get a lock without knowing how long you will hold it. If I remember right, the duration was around a minute while running the tests without the lock release part after each test.

lburgazzoli: For my understanding, what is the behavior if you call it with 0 as leaseId ?

jetcd-core/src/main/java/com/coreos/jetcd/data/ByteSequence.java
Grisu118: According to the [Javadoc](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/ByteString) ByteString is immutable like the String class. So a change should not be possible.

lburgazzoli: Ah correct, I got confused by some methods like concat

jetcd-core/src/main/java/com/coreos/jetcd/Lock.java
Grisu118: I did a short test with multiple runs with `leaseId` 0. 

```java
    for (int i = 0; i < 10; i++) {
      System.out.println("Start");
      CompletableFuture<LockResponse> feature = lockClient.lock(SAMPLE_NAME, 0);
      LockResponse response = feature.get();
      System.out.println("Got first lock: " + response.getKey().toStringUtf8());
      long time = System.currentTimeMillis();

      CompletableFuture<LockResponse> feature2 = lockClient.lock(SAMPLE_NAME, 0);
      LockResponse response2 = feature2.get();

      System.out.println("Got second lock (" + response2.getKey().toStringUtf8() + ") after " + (System.currentTimeMillis() - time) + "ms");

      lockClient.unlock(response2.getKey()).get();
    }
```

The second request returned always after 60 seconds.

lburgazzoli: So if we set it to 0 we don't have control over the timeout correct ?

Grisu118: Yes, I have no idea who defines the 60 seconds, found nothing about this in etcd docu

PR #269, opened 2017-12-29
Better OSGi support

lburgazzoli:
This PR is aimed to improve JEtcd OSGi support and includes the following changes to the current projects:

- there is a new artifact jetcd-common to have better separation of modules
- there is a new artifact jetcd-osgi that provides Apache Karaf features and some JEtcd services exposed via OSGi DS
- jetcd-all is now enabled and shades all the dependencies jetcd requires. this is needed because grpc has some split packages issues so it is not possible to deploy "plain" grpc artifacts in OSGi environments 

Some more work may be needed but I'd like to get this merged so people can test it.


fanminshi:
@lburgazzoli i took a quick glance of the code. it seems to me there are lot of code changes in different areas

* there is a new artifact jetcd-common to have better separation of modules
* there is a new artifact jetcd-osgi that provides Apache Karaf features and some JEtcd services exposed via OSGi DS
* jetcd-all is now enabled and shades all the dependencies jetcd requires. this is needed because   grpc has some split packages issues so it is not possible to deploy "plain" grpc artifacts in OSGi environments

is it possible to submit separate standalone pr for each? in that way, it can help better to understand all the changes happening in the pr.

lburgazzoli:
@fanminshi I can try to split it into different PRs but not sure if I can make it soon. 

Most of the code changes are just code relocation:

- **jetcd-common** at the moment contains only the etcd exceptions relocated from `jetcd-core/com.coreos.jetcd.exception`
- **jetcd-resolver** is the new home for `jetcd/com.coreos.jetcd.resolver`
- **jetcd-resolver-dbs-srv** is almost a rename of `jetcd-discovery-dns-srv` 
- **jetcd-osgi** is a brand new module with 2 small classes and related tests

The main changes apply to the POMs to properly configure OSGi stuffs and ion particular to **jetcd-all** which is now used as OSGi bundle as workaround for the following grpg-java issue with module systems:

- https://github.com/grpc/grpc-java/issues/3522
- https://github.com/grpc/grpc-java/issues/2847

jetcd-osgi/src/main/resources/features.xml
```diff
@@ -0,0 +1,89 @@
...
+
+-->
+<features xmlns="http://karaf.apache.org/xmlns/features/v1.4.0" name="coreos-jetcd">
+    <feature name="coreos-jetcd" description="coreos-jetcd" version="${project.version}}">
+
+        <details>CoreOS :: Java Client for etcd V3</details>
+
+        <feature prerequisite="true">wrap</feature>
+
+        <!-- google -->
```
fanminshi: why is this commented out? prefer codebase to be as minimum as possible.

jetcd-resolver-dns-srv/src/main/java/com/coreos/jetcd/resolver/dnssrv/DnsSrvUriResolver.java
```diff
@@ -50,7 +48,7 @@
     ENV.put("java.naming.provider.url", "dns:");
   }
 
-  private ConcurrentMap<String, EquivalentAddressGroup> cache;
+  private ConcurrentMap<String, SocketAddress> cache;
```
fanminshi: why changing `EquivalentAddressGroup` to `SocketAddress`?

jetcd-resolver/src/test/java/com/coreos/jetcd/resolver/NameResolverTest.java
```diff
@@ -31,7 +29,7 @@ public void testDefaults() throws Exception {
     final URIResolverLoader loader = URIResolverLoader.defaultLoader();
     final SmartNameResolver resolver = new SmartNameResolver("etcd", Collections.emptyList(), loader);
 
-    assertThat(resolver.getResolvers().stream().anyMatch(DirectUriResolver.class::isInstance)).isTrue();
+    Assertions.assertThat(resolver.getResolvers().stream().anyMatch(DirectUriResolver.class::isInstance)).isTrue();
```
fanminshi: any reason on changing imports to `import org.assertj.core.api.Assertions`?

fanminshi:
@lburgazzoli looks good in general. I quickly studied osgi and gain an general idea on your osgi implementation. it would be better if someone with more expertise can also review the osgi part of code. 

also the ci is broken. could you check upon that.

jetcd-resolver/src/test/java/com/coreos/jetcd/resolver/NameResolverTest.java
lburgazzoli: No done intentionally, guess the IDE did it while refactoring. Will fix it

jetcd-resolver-dns-srv/src/main/java/com/coreos/jetcd/resolver/dnssrv/DnsSrvUriResolver.java
lburgazzoli: To reduce packages inter dependencies in OSGi

jetcd-osgi/src/main/resources/features.xml
lburgazzoli: This is left as documentation to show how the karaf feature should looks like once grpc solves the split package issue

etc/scripts/run_etcd_docker.sh
```diff
@@ -87,7 +87,7 @@ docker run \
     --name etcd-ssl \
     --hostname etcd-ssl \
     --network etcd \
-    --volume $CERT_HOME:/etc/ssl/etcd \
+    --volume $CERT_HOME:/etc/ssl/etcd:Z \
```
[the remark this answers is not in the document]
lburgazzoli: Z is for SElinux, it should jave no effects on systems were SELinux is disabled

jetcd-osgi/src/main/resources/features.xml
fanminshi: i see. Could you add comment to explain the intention?

jetcd-resolver-dns-srv/src/main/java/com/coreos/jetcd/resolver/dnssrv/DnsSrvUriResolver.java
fanminshi: got it.

lburgazzoli:
@fanminshi I'm having some issues ot understand why the OSGi tests fails on travis-ci but not locally and I do not have time to digg into further so would it be ok if I mark as @Ignored OSGi tests for the moment ? I'd like to get this PR merged asap so it does not diverge too much from master.

fanminshi:
@lburgazzoli I understand your concern. I usually don't want to merge anything that has potential bug whether it is ci or code itself. Is there a rush in pushing osgi support in? afaik, not many changes are being pushed into jetcd. I think you are safe from worrying about pr diverging. Could we get another week or two of debugging. I might take a look as well. If we can't figure out by then, we can the pr with @ignored test option.

lburgazzoli:
Yeah I have no issue to wait some more weeks. In fact my main goal is to get jetcd in apache-camel and OSGi support is definitively something that would be more that welcome.

lburgazzoli:
@fanminshi I've rebased against latest master, can we get this merged ?

fanminshi:
yeah, let's get this merged.

TisonKun:
I found several diffs on changing `import io.etcd.jetcd.Auth;` to `import io.etcd.jetcd.api.Auth;`(e.g., `io.etcd.jetcd.auth.*Response`). However, the former is the proper one. Maybe it is done by automatically refactoring.

Issue #236, opened 2017-09-08
watch bug?


cc861010:
@org.junit.Test
    public void watch() throws Exception {

        Client client = null;
        Watch watch = null;
        Watch.Watcher watcher = null;
        try {
            client = Client.builder().endpoints("http://127.0.0.1:2379").build();
            watch = client.getWatchClient();
            watcher = watch.watch(ByteSequence.fromString("cc"));
            WatchResponse response = watcher.listen();
            for (WatchEvent event : response.getEvents()) {
                System.out.println(event.getEventType());
            }
        } catch (Exception e) {
            if (watcher != null) {
                watcher.close();
            }

            if (watch != null) {
                watch.close();
            }

            if (client != null) {
                client.close();
            }
        }

    }

my etcd is not running in http://127.0.0.1:2379 but it still can listen and never throw a exception

        
        <dependency>
            <groupId>com.coreos</groupId>
            <artifactId>jetcd-core</artifactId>
            <version>0.0.1</version>
        </dependency>

fanminshi:
@cc861010 ` client = Client.builder().endpoints("http://127.0.0.1:2379").build()`  never throw any connection error?

fanminshi:
@cc861010 another question, did you start the etcd server at all?

cc861010:
i mean the junit test run in an environment that does not have etcd server @fanminshi

fanminshi:
@cc861010 seems a bug to me. I would expect the client throw connection error.

fanminshi:
cc/ @skyao @俊雄 could you take a look on this issue.

ScienJus:
in `WatchImpl#processError`, if throwable is a `io.grpc.StatusRuntimeException: UNAVAILABLE: Transport closed for unknown reason`, we should throw it out?

fanminshi:
@ScienJus that's expected behavior. the watch should retry on temporary connection error. I think issue is that building `Client.builder().endpoints("http://127.0.0.1:2379").build()` should return an error but it doesn't.

ScienJus:
@fanminshi i see the https://github.com/coreos/etcd/tree/master/client also does not throw an error when building client. so i don't think it is a bug.

fanminshi:
[...]

PR #279, opened 2018-02-02
#236  watch: Notify watches of connection loss

Grisu118:
Always send an EtcdException with Unavailable Status to active watchers
on connection loss. This fix guards against concurrency issues when creating
watches while the connection to etcd is being restored.

### DIFF
--- a/jetcd-core/src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
+++ b/jetcd-core/src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
@@ -35,9 +35,11 @@ import com.coreos.jetcd.options.WatchOption;
 import com.coreos.jetcd.watch.WatchResponseWithError;
 import com.google.common.base.Strings;
 import com.google.protobuf.ByteString;
+
 import io.grpc.Status;
 import io.grpc.Status.Code;
 import io.grpc.stub.StreamObserver;
+
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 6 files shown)


fanminshi:
`Always send an EtcdException with Unavailable Status to active watchers
on connection loss.` I don't think the etcd go client has this behavior. that's why jetcd client doesn't throw EtcdException. But i need to double check on that.

fanminshi:
@Grisu118 have you read the go client watch code? the jetcd watch implementation matches of that.

jetcd-core/src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
```diff
@@ -94,54 +97,50 @@ public Watcher watch(ByteSequence key) {
...
 
     WatcherImpl watcher = new WatcherImpl(key, watchOption, this);
     this.pendingWatchers.add(watcher);
-
-    if (this.pendingWatchers.size() == 1) {
-      // head of the queue send watchCreate request.
-      WatchRequest request = this.toWatchCreateRequest(watcher);
-      this.getGrpcWatchStreamObserver().onNext(request);
-    }
+    this.triggerWatcherCreation();
```
fanminshi: what's the issue with previous implementation on 

```java
if (this.pendingWatchers.size() == 1) {		
 -      // head of the queue send watchCreate request.		
 -      WatchRequest request = this.toWatchCreateRequest(watcher);		
 -      this.getGrpcWatchStreamObserver().onNext(request);		
 -    }
```

Grisu118:
Sorry for the late answer. I saw issue #236 and the comment of @lburgazzoli. There he wrotes about returning from listen and resubscribe on next call of listen. As this behavior was helpful for our needs I tried to implement a solution. This pull request is the second attempt to do this, as my version had a bug. Need to do some internal testing to see if we can live without that behavior.

jetcd-core/src/main/java/com/coreos/jetcd/internal/impl/WatchImpl.java
Grisu118: the issue was that with the retry logic and multiple open watchers we got concurrency issues.

PR #295, opened 2018-03-03
Improve integration test framework

IgorPerikov:
according to this issue https://github.com/coreos/jetcd/issues/106 
this pull request 
1. introduce testcontainers framework usage for all suitable tests rewrite all tests, so they use their separate etcd clusters and nodes
2. remove old script
3. fixed travis build command (I hope :smile: ) so It doesn't launch old script anymore
4. fixed README section about launching tests


jetcd-core/src/test/java/com/coreos/jetcd/internal/impl/ClusterClientTest.java
```diff
@@ -20,29 +20,55 @@
...
    */
+
+  @BeforeTest
+  public void setUp() throws InterruptedException {
+    firstNode.start();
+    secondNode.start();
+    thirdNode.start();
+    endpoints = TestUtil.buildClientEndpoints(firstNode, secondNode, thirdNode);
+    peerUrls = TestUtil.buildPeerEndpoints(firstNode, secondNode, thirdNode);
+    TimeUnit.SECONDS.sleep(5);
```
IgorPerikov: Client was saying, that `cluster is unhealthy`. So I gave him some time here

fanminshi:
@IgorPerikov Thanks for the pr. I give this a thought. If we can wrap all the etcd containers  creation logic into a new class called `Cluster` can probably a better abstraction over just launch each node by itself. I am suggestion the following interface:

```java
/**
 * Cluster is an etcd cluster.
 */
public interface Cluster {

  // randClient randomly picks a etcd client within the cluster.
  Client randClient();

  // client gets the member client based on member idx within the cluster.
  Client client(int i);

  // terminate terminates the etcd cluster.
  void terminate();
}
```

Also a factory class to create it.
```java
public class ClusterFactory {
  // newCluster returns a launched cluster with client connection
  // for each cluster member.
  public static Cluster newCluster(int size) {
    return null;
  }
}
```

Let me know what you think.

cc/ @lburgazzoli

IgorPerikov:
I'm not exactly agree with `Cluster` interface, but it's a good idea to build another abstraction, which will handle all this infrastructure logic. I'll develop the concept and show it to you later

IgorPerikov:
@fanminshi I've pushed draft of the cluster abstraction. There might be still some small mistakes/bad namings or so, but general idea is there. More about some design decisions I've made:
1. Since many of the tests manipulate over endpoints themselves, I decided not to hide endpoints inside cluster abstraction, I don't think that rewriting logic of tests is good goal here, but eventually it might be a good idea to hide them too
2. As I can see there is no point in methods like `randClient()` and `client(int idx)`, because we operate with whole cluster anyway
3. Simple factory class/no fascinate builder etc: since all the tests requires either single node or 3 node cluster there is no big point in implementing extremely flexible method which can create custom size cluster, looks like overengineering for me
4. There is some custom annotations for test frameworks which will call `close()` methods on marked objects, but since it's not very common libraries I decided to kept the explicit `cluster.close()` calls in `@AfterXXX` methods

fanminshi:
`I'm not exactly agree with Cluster interface` sorry i wasn't being too clear on those. Those are copy/pasted from etcd's `Cluster`. It is okay that jetcd's abstraction is different.

`As I can see there is no point in methods like randClient() and client(int idx), because we operate with whole cluster anyway` Yeah, for now those are not useful. we don't have to include those.

jetcd-core/src/test/java/com/coreos/jetcd/internal/impl/KVTest.java
```diff
@@ -221,4 +227,9 @@ public void testNestedTxn() throws Exception {
     test.assertEquals(getResp2.getKvs().size(), 1);
     test.assertEquals(getResp2.getKvs().get(0).getValue().toStringUtf8(), oneTwoThree.toStringUtf8());
   }
+
+  @AfterTest
+  public void stop() throws IOException {
```
fanminshi: maybe use `public void tearDown()` instead? that is being used in everywhere else i think.

jetcd-core/src/test/java/com/coreos/jetcd/internal/impl/MaintenanceTest.java
```diff
@@ -117,4 +121,9 @@ public void testMoveLeader() throws ExecutionException, InterruptedException {
       client.getMaintenanceClient().moveLeader(followers.get(0)).get();
     }
   }
+
+  @AfterTest
+  public void stopContainers() throws IOException {
```
fanminshi: use `public void tearDown()`?

fanminshi:
@IgorPerikov the new interface looks great! just some nits then lgtm. thanks for the great work!

fanminshi:
also when you have addressed my comments/suggestions, could you also squash the commits? e.g 
a commit for updaing readme.md, a commit for update travis, a commit for adding `EtcdCluster` and its related stuff, a commit for updating tests using `EtcdCluster`, and etcd. I hope you get the idea.

fanminshi:
@IgorPerikov kindly ping?

IgorPerikov:
@fanminshi Hi! I had a day off, and I'm gonna do the rest of the pr-related work tonight, stay tune!

IgorPerikov:
@fanminshi done

PR #304, opened 2018-04-05
chore(build) configure version-maven-plugin to report updated dependencies

lburgazzoli:
When the build is executed with profile deps (-Pdeps) the
version-maven-plugin reports updated dependencies


fanminshi:
lgtm

PR #316, opened 2018-05-03
[API] Issue #278: Provide async methods for Watch

### LINKED ISSUE #278: Watch using CompletableFuture
Using watch in v0.0.1 blocks current thread, which makes it unusable in async environment. Making listen return CompletableFuture instead of blocking seems more appropriate.

sijie:
*Motiviation*

Currently Watch doesn't provide any async versions of methods to listen on watch responses. It also spawns a thread per watch, which is
not needed at all. The implementation can't be used in an environment that potentially set multiple watchers.

*Solution*

Introduce a `listenAsync` method and rewrite the implementation to avoid using a thread per watch.

This fixes #278

### DIFF
--- a/jetcd-core/src/main/java/com/coreos/jetcd/Watch.java
+++ b/jetcd-core/src/main/java/com/coreos/jetcd/Watch.java
@@ -16,6 +16,9 @@
 
 package com.coreos.jetcd;
 
+import static com.coreos.jetcd.common.exception.EtcdExceptionFactory.newClosedWatcherException;
+import static com.coreos.jetcd.common.exception.EtcdExceptionFactory.toEtcdException;
+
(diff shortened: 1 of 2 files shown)


sijie:
since the implementation is implementing the watch in async way and let sync method calling async, so existing watch unit tests already cover the code changes. so I didn't add more unit tests. but if there is a need, I am happy to add more tests.

lburgazzoli:
I'm re-writing Watch client so it is truly async so I'll put this PR on hold

kennylbj:
Any progress? Waiting for new release about async support @lburgazzoli

lburgazzoli:
@kennylbj need some more time

lburgazzoli:
should be ready by the end of the month

PR #363, opened 2018-08-20
enable more of error-prone's checks as errors to enforce them

vorburger:
see commit message for more details about changes made

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/internal/impl/LeaseImpl.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/internal/impl/LeaseImpl.java
@@ -21,6 +21,8 @@ import static io.etcd.jetcd.common.exception.EtcdExceptionFactory.newClosedLease
 import static io.etcd.jetcd.common.exception.EtcdExceptionFactory.newEtcdException;
 import static io.etcd.jetcd.common.exception.EtcdExceptionFactory.toEtcdException;
 
+import com.google.common.util.concurrent.ListeningScheduledExecutorService;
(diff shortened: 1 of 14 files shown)


jetcd-core/src/main/java/io/etcd/jetcd/internal/impl/LeaseImpl.java
```diff
@@ -41,16 +43,18 @@
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
```
lburgazzoli: I do prefer to use slf4j

vorburger: I actually personally do as well, but I had just copy/pasted this blindly from elsewhere [this project seems to use JUL so far](https://github.com/etcd-io/jetcd/search?q=java.util.logging&unscoped_q=java.util.logging) (see link) ... but I [now see that elsewhere the project also uses slf4j](https://github.com/etcd-io/jetcd/search?q=slf4j&unscoped_q=slf4j) - hm, this is not great. Let's first clean this up in #370, and then I change this?

lburgazzoli:
@vorburger can you review conflicts ?

vorburger:
> @vorburger can you review conflicts ?

done, also using SLF4j instead of JUL in new classes now (following #370), have a look if OK you.

PR #394, opened 2018-09-10
make Apache Commons IO a scope test dependency of jetcd-core

vorburger:
so that it does not get included in the (fat) jetcd-all, without shading, which is bad.  Stumbled upon this while looking into jetcd-all for #393 (but this DOES NOT fix that issue).


xiang90:
lgtm

PR #437, opened 2018-11-07
Fix compaction exception and save revision on progress notify

daMupfel:
Hi we recently enabled compaction in our etcd cluster and run into some issues.

From our tests the compactrevision != 0 case happens only if the canceled flag is also set.
I wrote a test to confirm that behavior (although i don't know if the other one is possible).

The second thing i changed is that the revision is saved on progress notify. This was in the last implementation of the watch and got removed, i am not sure why. (The go implementation also seems to save nextrev if it is a progress notifiy message).

Please feel free to comment :)

Thanks in advance,

David

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/WatchImpl.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/WatchImpl.java
@@ -195,7 +195,9 @@ final class WatchImpl implements Watch {
         String reason = response.getCancelReason();
         Throwable error;
 
-        if (Strings.isNullOrEmpty(reason)) {
+        if (response.getCompactRevision() != 0) {
+          error = newCompactedException(response.getCompactRevision());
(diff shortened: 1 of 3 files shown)


jetcd-core/src/main/java/io/etcd/jetcd/WatchImpl.java
```diff
@@ -195,7 +195,9 @@ public void onNext(WatchResponse response) {
         String reason = response.getCancelReason();
         Throwable error;
 
-        if (Strings.isNullOrEmpty(reason)) {
+        if (response.getCompactRevision() != 0) {
```
daMupfel: This could also be done like this:
Leave old cold and add the following:

`} else if (response.getCanceled() && response.getCompactRevision() == 0) {`

lburgazzoli:
thx

PR #438, opened 2018-11-07
use @findbugs.NonNull annotation from SpotBugs instead of javax (#436)

vorburger:
because of that whole (old) JSR 305 controversy; see e.g.

- https://github.com/findbugsproject/findbugs/issues/88
- https://github.com/spotbugs/spotbugs/issues/130
- https://github.com/spotbugs/spotbugs/pull/180

### DIFF
--- a/jetcd-launcher/pom.xml
+++ b/jetcd-launcher/pom.xml
@@ -36,15 +36,13 @@
 
     <dependencies>
         <dependency>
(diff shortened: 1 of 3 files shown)


vorburger:
@lburgazzoli this is a follow up to #435 and instead of #436 ...

PR #439, opened 2018-11-08
jetcd-core(watch): fix resuming watch after server disconnection

fdimuccio:
Hi,
This PR should address stale watchers due to etcd node restarts. 

The watcher stream was not completed after receiving UNAVAILABLE or INTERNAL errors from the etcd cluster making the listener attached to it stale. The reason is that the subsequent call to WatcherImpl::resume does nothing unless the stream is set to null. This patchset completes the stream and sets it to null regardless of the error, in this way the call to resume will always recreate a new stream.

Feel free to comment and let me know if my assumptions are incorrect.


lburgazzoli:
looks correct, is it possible to create a test for that ?

fdimuccio:
I added a test in the referenced PR #443

PR #495, opened 2019-02-06
add tests to illustrate how to use EtcdClusterFactory

vorburger:
The EtcdClusterUsingTest unfortunately cannot be put into
jetcd-launcher/src/test/java, because it requires io.etcd.jetcd (our
client, which is in core), so we put that one in core but keep at least
the more basic EtcdClusterStartTest into launcher.

### DIFF
--- a/jetcd-core/pom.xml
+++ b/jetcd-core/pom.xml
@@ -130,6 +130,11 @@
             <artifactId>commons-io</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.truth</groupId>
+            <artifactId>truth</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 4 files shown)


fanminshi:
@vorburger isn't the EtcdClusterFactory used mainly for testing? like in the BeforeTest and AfterTest methods. I think it might be better if we add documentation on how to use the EtcdFactory in for testing jetcd. Related to https://github.com/etcd-io/jetcd/issues/506.

vorburger:
> @vorburger isn't the EtcdClusterFactory used mainly for testing? 

@fanminshi yes, it is; and this PR just proposed a usage example.

> I think it might be better if we add documentation on how to use the EtcdFactory in for testing jetcd.

The README (now) [actually already documents](https://github.com/etcd-io/jetcd#launcher) the recommend use of the `@Rule  EtcdClusterResource`.

This PR is already part of #498 so when that one is merged, we can close this one.

vorburger:
> This PR is already part of #498 so when that one is merged, we can close this one.

Abandoning this PR as this commit is now on master as part of the meanwhile merged #498.

PR #503, opened 2019-02-11
Support namespace in jetcd

### LINKED ISSUE #472: Support namespace in jetcd
Previously client API in golang has already supported [namespace](https://godoc.org/github.com/coreos/etcd/clientv3/namespace), it'll be quite useful if jetcd also support namespace since it's error-prone to concatenate a namespace prefix with each key before KV/Lock/Watch operations.

MalcolmSanders:
This PR enables users to specify a namespace for each key used in jetcd client.

Fixes #472


fanminshi:
if you can format the commit title to follow this https://github.com/etcd-io/jetcd/blob/master/CONTRIBUTING.md#format-of-the-commit-message, that will be great.

i'll take a look of this pr tonight.

MalcolmSanders:
@fanminshi Thanks for your advise, I've created a new PR #504 and this will be abandoned.

Issue #309, opened 2018-04-10
Cleanup tests : remove TestNG


lburgazzoli:
We have some tests using TestNG an some others using JUnit, we need to get rid of TestNG

liukun4515:
could you assign this issue to me？

lburgazzoli:
I can't but you are very welcome to submit a pr

fanminshi:
@liukun4515 just do it.

liukun4515:
ok

lburgazzoli:
@liukun4515 I think every unit test should not depend on any other test so you should set-up etcd according to your need in your test

fanminshi:
`The JUnit doesn't support the dependency testing.`
I don't think we should have any dependency testing at all. it is generally a bad practice in testing. We want each test to be independent from each other.

lburgazzoli:
@liukun4515 have you been able to progress on this issue ?

vorburger:
I've made a start with this issue in #358; that's 1/4 done. The `ClusterClientTest`, `AuthClientTest` and `LockTest` still have to be done (hopefully by others than me) before we can fully remove the testng dependency.

The point re. dependency testing is easy enough to resolve - e.g. in #358 the testDelete() now simply first calls testPut() instead of TestNG's `dependsOnMethods`. That's simpler, clearer and better than (ab)using JUnit's  `@FixMethodOrder` to replace `dependsOnMethods`, IMHO.

lburgazzoli:
@vorburger yes dependsOnMethods and other similar stuffs should be removed, we can also have independant test so a change on one of the methods does not imact all the test.

pdfkpb:
Hello there, I'm working on cleaning up the TestNG dependency, and I'm interested if we'd also like to maintain the dependency on AssertJ, it does seem useful, but if we don't want it, then I can clean it up at the same time.

vorburger:
@pdfkpb whatever works best for you. Perhaps two separate commits & PRs is simpler to review?

PS: Personally I anyway prefer [G's Truth](https://github.com/google/truth) over AssertJ nowadays, but that's not this issue.

lburgazzoli:
At the moment I would focus to get rid of TestNG only

lburgazzoli:
@pdfkpb any update ?

pdfkpb:
I'm coming to the end of an internship so it's been a bit busy here, but I'll see about getting a pr in by the end of next week.

[...]

PR #504, opened 2019-02-12
jetcd-core: support namespace in jetcd client

### LINKED ISSUE #472: Support namespace in jetcd

MalcolmSanders:
This enables users to specify a namespace when building a jetcd client.

Fixes #472

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/ClientBuilder.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/ClientBuilder.java
@@ -57,6 +57,7 @@ public final class ClientBuilder implements Cloneable {
   private Integer maxInboundMessageSize;
   private Map<Metadata.Key, Object> headers;
   private List<ClientInterceptor> interceptors;
+  private ByteSequence namespace = ByteSequence.EMPTY;
 
   ClientBuilder() {
   }
@@ -139,6 +140,24 @@ public final class ClientBuilder implements Cloneable {
   public ClientBuilder password(ByteSequence password) {
     checkNotNull(password, "password can't be null");
     this.password = password;
+    return this;
+  }
+
+  public ByteSequence namespace() {
+    return namespace;
+  }
+
+  /**
+   * config the namespace of keys used in {@code KV}, {@code Txn}, {@code Lock} and {@code Watch}.
+   * "/" will be treated as no namespace.
+   *
+   * @param namespace the namespace of each key used
+   * @return this builder
+   * @throws NullPointerException if namespace is <code>null</code>
+   */
+  public ClientBuilder namespace(ByteSequence namespace) {
+    checkNotNull(namespace, "namespace can't be null");
+    this.namespace = namespace;
     return this;
   }
 
--- a/jetcd-core/src/main/java/io/etcd/jetcd/kv/TxnResponse.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/kv/TxnResponse.java
@@ -22,6 +22,8 @@ import static io.etcd.jetcd.api.ResponseOp.ResponseCase.RESPONSE_RANGE;
 import static io.etcd.jetcd.api.ResponseOp.ResponseCase.RESPONSE_TXN;
 
 import io.etcd.jetcd.AbstractResponse;
+import io.etcd.jetcd.ByteSequence;
+
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -31,15 +33,17 @@ import java.util.stream.Collectors;
  */
 public class TxnResponse extends AbstractResponse<io.etcd.jetcd.api.TxnResponse> {
 
-  // TODO add txnResponsesRef when nested txn is implemented.
+  private final ByteSequence namespace;
+
   private List<PutResponse> putResponses;
   private List<GetResponse> getResponses;
   private List<DeleteResponse> deleteResponses;
   private List<TxnResponse> txnResponses;
 
 
-  public TxnResponse(io.etcd.jetcd.api.TxnResponse txnResponse) {
+  public TxnResponse(io.etcd.jetcd.api.TxnResponse txnResponse, ByteSequence namespace) {
     super(txnResponse, txnResponse.getHeader());
+    this.namespace = namespace;
   }
 
   /**
@@ -56,7 +60,7 @@ public class TxnResponse extends AbstractResponse<io.etcd.jetcd.api.TxnResponse>
     if (deleteResponses == null) {
       deleteResponses = getResponse().getResponsesList().stream()
           .filter((responseOp) -> responseOp.getResponseCase() == RESPONSE_DELETE_RANGE)
-          .map(responseOp -> new DeleteResponse(responseOp.getResponseDeleteRange()))
+          .map(responseOp -> new DeleteResponse(responseOp.getResponseDeleteRange(), namespace))
           .collect(Collectors.toList());
     }
 
@@ -70,7 +74,7 @@ public class TxnResponse extends AbstractResponse<io.etcd.jetcd.api.TxnResponse>
     if (getResponses == null) {
       getResponses = getResponse().getResponsesList().stream()
           .filter((responseOp) -> responseOp.getResponseCase() == RESPONSE_RANGE)
-          .map(responseOp -> new GetResponse(responseOp.getResponseRange()))
+          .map(responseOp -> new GetResponse(responseOp.getResponseRange(), namespace))
           .collect(Collectors.toList());
     }
 
@@ -84,7 +88,7 @@ public class TxnResponse extends AbstractResponse<io.etcd.jetcd.api.TxnResponse>
     if (putResponses == null) {
       putResponses = getResponse().getResponsesList().stream()
           .filter((responseOp) -> responseOp.getResponseCase() == RESPONSE_PUT)
-          .map(responseOp -> new PutResponse(responseOp.getResponsePut()))
+          .map(responseOp -> new PutResponse(responseOp.getResponsePut(), namespace))
           .collect(Collectors.toList());
     }
 
@@ -95,7 +99,7 @@ public class TxnResponse extends AbstractResponse<io.etcd.jetcd.api.TxnResponse>
     if (txnResponses == null) {
       txnResponses = getResponse().getResponsesList().stream()
           .filter((responseOp) -> responseOp.getResponseCase() == RESPONSE_TXN)
-          .map(responseOp -> new TxnResponse(responseOp.getResponseTxn()))
+          .map(responseOp -> new TxnResponse(responseOp.getResponseTxn(), namespace))
           .collect(Collectors.toList());
     }
 
--- a/jetcd-core/src/test/java/io/etcd/jetcd/ClientBuilderTest.java
+++ b/jetcd-core/src/test/java/io/etcd/jetcd/ClientBuilderTest.java
@@ -16,33 +16,42 @@
 package io.etcd.jetcd;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
+import com.google.common.base.Charsets;
 import io.grpc.netty.NettyChannelBuilder;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Random;
-import org.junit.Test;
+import java.util.stream.Stream;
+
 
 public class ClientBuilderTest {
 
-  @Test(expected = NullPointerException.class)
+  @Test
   public void testEndPoints_Null() {
-    Client.builder().endpoints((URI)null);
+    assertThrows(NullPointerException.class, () -> Client.builder().endpoints((URI)null));
   }
 
-  @Test(expected = IllegalArgumentException.class)
+  @Test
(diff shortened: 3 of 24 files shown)


fanminshi:
I am just curious if you have taken a look of how etcd does the namspacing https://github.com/etcd-io/etcd/tree/master/clientv3/namespace?

MalcolmSanders:
> I am just curious if you have taken a look of how etcd does the namspacing https://github.com/etcd-io/etcd/tree/master/clientv3/namespace?

Yes. The implementation in etcd just adds one wrapper class for each client (KV, lease, watch) with necessary prefix and unprefix operations, while I just add prefix and unprefix operations inside the implementation of original clients. I think both ways are okay. If you have a preference on that, I'll change it to wrapper class way, it won't take much efforts.

I guess you might be concerned with the modification on Response classes which I added the namespace to their constructors in order to do unprefix operations, for example, the constructor of io.etcd.jetcd.kv.PutResponse is changed
from
`public PutResponse(io.etcd.jetcd.api.PutResponse putResponse)`
To
`public PutResponse(io.etcd.jetcd.api.PutResponse putResponse, ByteSequence namespace)`

Since in jetcd "io.etcd.jetcd.api.PutResponse" is a pb structure (code generated) and keys are stored as ByteString inside pb structures, it seems difficult or hack to tamper contents in pb structures. While keys of responses in etcd are easy to be tampered because they are stored in byte arrays and not code generated. Do you have any suggestion on it ?

jetcd-core/src/test/java/io/etcd/jetcd/ClientBuilderTest.java
```diff
@@ -53,4 +58,45 @@ public void testMaxInboundMessageSize() throws URISyntaxException {
...
+            ByteSequence.from("/namespace1/", Charsets.UTF_8)),
+        new TestCase(ByteSequence.from("namespace2/", Charsets.UTF_8),
+            ByteSequence.from("namespace2/", Charsets.UTF_8))
+    );
+    testCases.forEach(testCase -> {
+      try {
+        final ClientBuilder builder =  Client.builder().endpoints(new URI("http://127.0.0.1:2379"))
+            .namespace(testCase.namespace);
+        final ClientConnectionManager connectionManager = new ClientConnectionManager(builder);
+        assertEquals(testCase.expectedNamespace, connectionManager.getNamespace());
```
fanminshi: could you use `assertThat(actual).isEqualTo(expected)` from assertJ instead of assertEquals? I think it is best to have a consistent testing tool though out the codebase.

jetcd-core/src/test/java/io/etcd/jetcd/ClientBuilderTest.java
```diff
@@ -53,4 +58,45 @@ public void testMaxInboundMessageSize() throws URISyntaxException {
...
+      public ByteSequence expectedNamespace;
+
+      public TestCase(ByteSequence namespace, ByteSequence expectedNamespace) {
+        this.namespace = namespace;
+        this.expectedNamespace = expectedNamespace;
+      }
+    }
+    List<TestCase> testCases = Arrays.asList(
+        // namespace setting and expected namespace
+        new TestCase(ByteSequence.EMPTY, ByteSequence.EMPTY),
```
fanminshi: Do you think you can use junit Parameterized tests here instead. I saw you use it in the later pr.

jetcd-core/src/test/java/io/etcd/jetcd/KVNamespaceTest.java
```diff
@@ -0,0 +1,500 @@
+/**
+ * Copyright 2017 The jetcd authors
```
fanminshi: Copyright 2019?

jetcd-core/src/test/java/io/etcd/jetcd/KVNamespaceTest.java
```diff
@@ -0,0 +1,500 @@
...
+      public TestCase(ByteSequence namespace, ByteString key, ByteString end,
+          ByteString wKey, ByteString wEnd) {
+        this.namespace = namespace;
+        this.key = key;
+        this.end = end;
+        this.wKey = wKey;
+        this.wEnd = wEnd;
+      }
+    }
+    List<TestCase> testCases = Arrays.asList(
```
fanminshi: I know you can do similar testing by different cases using dataprovider but i don't want to testNG in jetcd and actually want to get rid of it. as you can see jetcd testing has a mix of junit and testNG. it gets confusing. Is it possible to do the same thing using junit's Parameterized tests here?

jetcd-core/src/test/java/io/etcd/jetcd/KVNamespaceTest.java
```diff
@@ -0,0 +1,500 @@
...
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+import static org.junit.Assert.assertEquals;
```
fanminshi: prefer to use assertJ's assertThat instead of junit.Assert.assertEquals in order to be consistent of major of other tests.

fanminshi:
`I guess you might be concerned with the modification on Response classes which I added the namespace to their constructors in order to do unprefix operations`
It seems to me that users never creates Response object by themselves so I think it is totally fine to change the constructor. 

`Yes. The implementation in etcd just adds one wrapper class for each client (KV, lease, watch) with necessary prefix and unprefix operations, while I just add prefix and unprefix operations inside the implementation of original clients.`
Very Nice that you have taken a look of the etcdcode base. I saw some similarity in this PR. thats why I asked :). Also I think the current approach that you have seems very intuitive which is also similar to how curator framework does namespacing. 

One last thing, you probably can tell our testing code is mixed in different testing style and testing framework which is quite confusing. I am trying to limit the testing codebase to use only junit and assertJ for any newly added tests. related issue https://github.com/etcd-io/jetcd/issues/309 cc/ @lburgazzoli @vorburger. 

Overall the interface and approach looks good. However, I still need to read the implementation in details. So more reviews coming up in the fews days.

Thanks for the PR.

MalcolmSanders:
Thanks for replying.

> One last thing, you probably can tell our testing code is mixed in different testing style and testing framework which is quite confusing. I am trying to limit the testing codebase to use only junit and assertJ for any newly added tests. 

I agree with you. Actually I myself also got confused by these two different test frameworks used in tests. I'll remove testNG appeared in this PR.

jetcd-core/src/test/java/io/etcd/jetcd/KVNamespaceTest.java
MalcolmSanders: As to parameterized tests, there is a shortcoming of parameterized test in junit4:
we can only do parameterize test per test class. That is to say, if we configure two sets of parameters, then all the cases in this test class will be executed twice. While there should be some situations where not all the tests need to be parameterized, furthermore maybe not all the cases use the same parameter source.

A possible solution is to divide cases into test classes by their parameter sets, which might lose readability of the integral test design of a module.

In junit5 parameter source is introduced, while testNG also provides similar api. Since we decided to remove testNG, I wonder if we could change junit from junit4 to junit5 ?

Ref:
https://blog.parasoft.com/how-to-create-junit-parameterized-tests-faster
https://github.com/junit-team/junit4/wiki/Parameterized-tests
https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests

lburgazzoli: +1 for junit 5

fanminshi: Fine with me to use junit 5

jetcd-core/src/test/java/io/etcd/jetcd/ClientBuilderTest.java
MalcolmSanders: done

jetcd-core/src/test/java/io/etcd/jetcd/ClientBuilderTest.java
MalcolmSanders: done

jetcd-core/src/test/java/io/etcd/jetcd/KVNamespaceTest.java
MalcolmSanders: It has to be "Copyright 2017" since it's a rule listed in checkstyle-checker.xml

jetcd-core/src/test/java/io/etcd/jetcd/KVNamespaceTest.java
MalcolmSanders: done

jetcd-core/src/test/java/io/etcd/jetcd/KVNamespaceTest.java
fanminshi: that's right. we need to update that in the checkstyle-checker.xml. cc/ @lburgazzoli

jetcd-core/src/main/java/io/etcd/jetcd/ClientBuilder.java
```diff
@@ -142,6 +143,36 @@ public ClientBuilder password(ByteSequence password) {
...
+   * config the namespace of keys used in {@code KV}, {@code Txn}, {@code Lock} and {@code Watch}.
+   * "/" will be treated as no namespace.
+   *
+   * @param namespace the namespace of each key used
+   * @return this builder
+   * @throws NullPointerException if namespace is <code>null</code>
+   */
+  public ClientBuilder namespace(ByteSequence namespace) {
+    checkNotNull(namespace);
+    if (namespace.getByteString().isEmpty()) {
```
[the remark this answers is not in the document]
MalcolmSanders: yes, done.

jetcd-core/src/main/java/io/etcd/jetcd/kv/TxnResponse.java
```diff
@@ -31,15 +34,22 @@
...
   private List<TxnResponse> txnResponses;
 
 
-  public TxnResponse(io.etcd.jetcd.api.TxnResponse txnResponse) {
+  public TxnResponse(io.etcd.jetcd.api.TxnResponse txnResponse, ByteSequence namespace) {
     super(txnResponse, txnResponse.getHeader());
+    this.namespace = namespace;
+  }
+
+  @VisibleForTesting
```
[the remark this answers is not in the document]
MalcolmSanders: yes, actually the second constructor is not needed. I removed it in the latest version.

jetcd-core/src/main/java/io/etcd/jetcd/ClientBuilder.java
```diff
@@ -142,6 +143,36 @@ public ClientBuilder password(ByteSequence password) {
...
+    } else if (namespace.size() == 1
+        && namespace.getByteString().equals(ByteSequence.NAMESPACE_DELIMITER.getByteString())) {
+      // treat "/" as no namespace
+      this.namespace = ByteSequence.EMPTY;
+    } else if (namespace.getByteString().byteAt(namespace.getByteString().size() - 1) == '/') {
+      this.namespace = namespace;
+    } else {
+      // make sure the namespace ends with '/'
+      this.namespace = ByteSequence.from(namespace.getByteString().concat(
+          ByteSequence.NAMESPACE_DELIMITER.getByteString()));
```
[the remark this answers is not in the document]
MalcolmSanders: I think you might miss the previous condition, if the namespace ends with '/', do nothing.
`    } else if (namespace.getByteString().byteAt(namespace.getByteString().size() - 1) == '/') {
      this.namespace = namespace;
    }`

> Also I am unsure if we should enforce the ending the namespace with '/' since etcd go client doesn't do it if I recall correctly.
Yes, I've noticed that.In etcd, suppose a given key a, etcd will treat a+1 as the excluded end range.
I did it on purpose since I though it might be error-prone for users if we don't enforce the ending character. Suppose there are two namespace 'a' and 'aa':
We will get all the keys inside namespace 'aa' once we try to list all the keys inside namespace 'a' if we don't ending the ending '/'.
What do you think of it ?

fanminshi:
looks good in general.

jetcd-core/src/main/java/io/etcd/jetcd/ClientBuilder.java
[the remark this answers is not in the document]
fanminshi: That's good edge case. However, I don't think adding '/' will solve the issue. Suppose that there are two namespaces `a/` and `a//`. what if I do a range GET for all the keys with the name space `a/` and with the range end be`a/ + 1` , I would also get all the keys from `a//` right?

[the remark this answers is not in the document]
fanminshi: I think for now we should stick with etcd implementation and file an issue on etcd to discuss the edge case that you were thinking about. Once we reach a  conclusion, then we can have another pr to update jetcd.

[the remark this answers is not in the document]
MalcolmSanders: Got it, I'll change this part in the next version.

MalcolmSanders:
Previously all the comments are resolved. Just a friendly reminder~ @fanminshi @lburgazzoli

fanminshi:
lgtm thanks for contributing!

Issue #526, opened 2019-03-11
Fails to run test on master branch


fanminshi:
```
$ ./mvnw test
[INFO] Scanning for projects...
[INFO] ------------------------------------------------------------------------
[INFO] Detecting the operating system and CPU architecture
[INFO] ------------------------------------------------------------------------
[INFO] os.detected.name: osx
[INFO] os.detected.arch: x86_64
[INFO] os.detected.version: 10.14
[INFO] os.detected.version.major: 10
[INFO] os.detected.version.minor: 14
[INFO] os.detected.classifier: osx-x86_64
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Build Order:
[INFO]
[INFO] jetcd                                                              [pom]
[INFO] jetcd-launcher                                                     [jar]
[INFO] jetcd-launcher-maven-plugin                               [maven-plugin]
[INFO] jetcd-common                                                    [bundle]
[INFO] jetcd-resolver                                                  [bundle]
[INFO] jetcd-core                                                      [bundle]
[INFO] jetcd-launcher-maven-plugin-test                                   [jar]
[INFO] jetcd-resolver-dns-srv                                          [bundle]
[INFO] jetcd-extensions                                                [bundle]
[INFO] jetcd-all                                                       [bundle]
[INFO] jetcd-osgi-parent                                                  [pom]
[INFO] jetcd-osgi                                                      [bundle]
[INFO] jetcd-karaf                                                     [bundle]
[INFO] jetcd-examples                                                     [pom]
[INFO] jetcd-watch                                                        [jar]
[INFO] jetcd-ctl                                                          [jar]
[INFO]
[INFO] ------------------------< io.etcd:jetcd-parent >------------------------
[INFO] Building jetcd 0.3.1-SNAPSHOT                                     [1/16]
[INFO] --------------------------------[ pom ]---------------------------------
[INFO]
[INFO] --- directory-maven-plugin:0.3.1:highest-basedir (directories) @ jetcd-parent ---
[INFO] Highest basedir set to: /Users/fanminshi/work/src/github.com/fanminshi/jetcd
[INFO]
[INFO] -----------------------< io.etcd:jetcd-launcher >-----------------------
[INFO] Building jetcd-launcher 0.3.1-SNAPSHOT                            [2/16]
[INFO] --------------------------------[ jar ]---------------------------------
[INFO]
[INFO] --- maven-checkstyle-plugin:3.0.0:check (validate) @ jetcd-launcher ---
[...]

PR #533, opened 2019-03-16
pom: disbale error-prone

fanminshi:
error-prone plugin is causing jetcd unable to compile ref:https://github.com/etcd-io/jetcd/issues/526.
let's disable it for now.


vorburger:
@fanminshi this seems a little "drastic" :smile_cat:  to me .. perhaps you can figure out #526 instead?

fanminshi:
@vorburger  agreed!

PR #543, opened 2019-03-20
jetcd-core: remove testNG, introduce "bytesOf"

### LINKED ISSUE #309: Cleanup tests : remove TestNG

yb172:
TestNG is removed to unify stack used for testing, `ByteSequence.from(<str>, UTF_8)` is replaced with more concise `bytesOf(<str>)`

Fixes #309

### DIFF
--- a/jetcd-core/src/test/java/io/etcd/jetcd/AuthClientTest.java
+++ b/jetcd-core/src/test/java/io/etcd/jetcd/AuthClientTest.java
@@ -15,262 +15,158 @@
  */
 package io.etcd.jetcd;
 
-import static com.google.common.base.Charsets.UTF_8;
+import static io.etcd.jetcd.TestUtil.bytesOf;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
-import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.assertj.core.api.Assertions.assertThat;
 
 import io.etcd.jetcd.auth.AuthRoleGetResponse;
 import io.etcd.jetcd.auth.AuthRoleListResponse;
 import io.etcd.jetcd.auth.Permission;
 import io.etcd.jetcd.auth.Permission.Type;
-import io.etcd.jetcd.launcher.EtcdCluster;
-import io.etcd.jetcd.launcher.EtcdClusterFactory;
-import java.io.IOException;
+import io.etcd.jetcd.launcher.junit.EtcdClusterResource;
+
 import java.net.URI;
 import java.util.List;
-import java.util.concurrent.ExecutionException;
-import org.testng.annotations.AfterTest;
-import org.testng.annotations.BeforeTest;
-import org.testng.annotations.Test;
+
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
 
 /**
  * test etcd auth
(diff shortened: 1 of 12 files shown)


yb172:
@fanminshi @lburgazzoli Please take a look

jetcd-core/src/test/java/io/etcd/jetcd/AuthClientTest.java
```diff
@@ -15,262 +15,158 @@
...
+
 import java.net.URI;
 import java.util.List;
-import java.util.concurrent.ExecutionException;
-import org.testng.annotations.AfterTest;
-import org.testng.annotations.BeforeTest;
-import org.testng.annotations.Test;
+
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
```
fanminshi: beforeClass and Classrule is a junit 4 thing right? I think we currently want to move to junit 5. So maybe this refactoring can leverage junit 5 syntax?

This test https://github.com/etcd-io/jetcd/blob/master/jetcd-core/src/test/java/io/etcd/jetcd/KVNamespaceTest.java uses junit 5. Maybe you can base it off that or come with something better>

fanminshi:
overall the pr looks good in general.

jetcd-core/src/test/java/io/etcd/jetcd/AuthClientTest.java
yb172: It is JUnit4 thing and I tried to refactor everything to JUnit5 but wasn't able to do so. I haven't found an easy way to convert `@Rule` to JUnit5 extension. As you mentioned before this PR already does 2 things so I didn't want to add one more.

jetcd-core/src/test/java/io/etcd/jetcd/AuthClientTest.java
```diff
@@ -15,262 +15,158 @@
...
-import java.io.IOException;
+import io.etcd.jetcd.launcher.junit.EtcdClusterResource;
+
 import java.net.URI;
 import java.util.List;
-import java.util.concurrent.ExecutionException;
-import org.testng.annotations.AfterTest;
-import org.testng.annotations.BeforeTest;
-import org.testng.annotations.Test;
+
```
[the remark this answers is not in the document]
yb172: I would vote for whatever is supported by most of IDEs by default. For some reason if I tick "optimize on the fly" in IDEA it messes up imports so checkstyle complains.
Let's consider this in a separate issue.

jetcd-core/src/test/java/io/etcd/jetcd/AuthClientTest.java
fanminshi: fair enough, would you like to convert this Junit 5 in another pr then?

fanminshi:
lgtm defer to @lburgazzoli

jetcd-core/src/test/java/io/etcd/jetcd/AuthClientTest.java
yb172: I would take another look on how to replace `@Rule`. If I would find a way - yes, I'll do another PR, if not - then I won't :)

PR #555, opened 2019-03-26
550: all: enable checkstyle for tests and fix style errors

### LINKED ISSUE #550: Check targets appear to be broken when running locally
When running `./mvnw checkstyle:check` task reports success while it fails on CI because of unused import.

Also when running `./mvnw license:check` following error message is printed:

```text
[ERROR] Failed to execute goal com.mycila:license-maven-plugin:3.0:check (default-cli) on project jetcd-parent: Execution default-cli of goal com.mycila:license-maven-plugin:3.0:check failed: Cannot read header document ${jetcd.project.root}/etc/license.txt. Cause: Resource ${jetcd.project.root}/etc/license.txt not found in file system, classpath or URL: no protocol: ${jetcd.project.root}/etc/license.txt -> [Help 1]
```

yb172:
For better consistency enable checkstyle for tests and reformat tests to comply with rules.

Fixes #550

### DIFF
--- a/jetcd-core/src/test/java/io/etcd/jetcd/WatchUnitTest.java
+++ b/jetcd-core/src/test/java/io/etcd/jetcd/WatchUnitTest.java
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package io.etcd.jetcd;
 
 import static io.etcd.jetcd.TestUtil.bytesOf;
@@ -54,16 +55,13 @@ import org.mockito.ArgumentMatcher;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
-/**
- * watch test case.
- */
 // TODO(#548): Add global timeout for tests once JUnit5 supports it
 @ExtendWith(MockitoExtension.class)
 // TODO(#549): Remove GrpcServerRule and remove this annotation
 @EnableRuleMigrationSupport
 public class WatchUnitTest {
 
-  private final static ByteSequence KEY = bytesOf("test_key");
+  private static final ByteSequence KEY = bytesOf("test_key");
   @Rule
   public final GrpcServerRule grpcServerRule = new GrpcServerRule().directExecutor();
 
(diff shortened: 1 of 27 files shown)


pom.xml
```diff
@@ -708,7 +710,7 @@
                 <artifactId>license-maven-plugin</artifactId>
                 <version>${license-maven-plugin.version}</version>
                 <configuration>
-                    <header>${jetcd.project.root}/etc/license.txt</header>
+                    <header>${session.executionRootDirectory}/etc/license.txt</header>
```
yb172: @lburgazzoli does that look ok to you? I'm not an expert in maven, but with this change `./mvnw license:check` works locally for me.

yb172:
@fanminshi @lburgazzoli Please take a look

pom.xml
lburgazzoli: it does not look correct as that variable points to the path where the maven command was executed so if I go to a submodule, then the license check would not work.

That is the reason why the property jetcd.project.root exists, we should investigate why it didn’t get resolved.

yb172: All right, thank you for explanation. I would revert it back then.

jetcd-core/src/test/java/io/etcd/jetcd/WatchUnitTest.java
```diff
@@ -321,28 +320,28 @@ public void testWatcherCreateOnCancellationWithNoReason() throws InterruptedExce
...
 
     try (Watch.Watcher watcher = watchClient.watch(KEY, listener)) {
       WatchResponse createdResponse = createWatchResponse(0);
       responseObserverRef.get().onNext(createdResponse);
 
       WatchResponse canceledResponse = WatchResponse
-              .newBuilder()
-              .setCanceled(true)
-              .build();
+          .newBuilder()
```
fanminshi: is this auto formatted? not sure why it has 4 spaces as the indent.

jetcd-core/src/test/java/io/etcd/jetcd/WatchUnitTest.java
```diff
@@ -351,29 +350,29 @@ public void testWatcherCreateOnCancellationWithReason() throws InterruptedExcept
     CountDownLatch latch = new CountDownLatch(1);
     AtomicReference<Throwable> ref = new AtomicReference<>();
     Watch.Listener listener = Watch.listener(
-      r -> {},
-      t -> {
-        ref.set(t);
-        latch.countDown();
-      }
+        r -> { },
```
fanminshi: here too.

fanminshi:
LGTM if all the changes are according to checkstyle and google java style sheet.

jetcd-core/src/test/java/io/etcd/jetcd/WatchUnitTest.java
yb172: It's formatted according to styleguide. Weird thing is that linter complains if indentation is less that what it should be, but not when it's more.

fanminshi: i see.

PR #563, opened 2019-04-05
README: update ByteSequence usage in README.md

husterxsp:
Currently, ByteSequence don't have fromString and replace it with from, so just update readme usage~


fanminshi:
could you change the commit title to be `README: update ByteSequence usage in README.md`
Then lgtm

husterxsp:
OK. It has been changed.

PR #604, opened 2019-07-24
cmp.Op: Add NOT_EQUAL operator

### LINKED ISSUE #558: Implement Compare.CompareResult.NOT_EQUAL
It looks like Cmp.Op does not implement NOT_EQUAL, but NOT_EQUAL is in the proto specs.

actien:
adds the NOT_EQUAL operator, which is defined in the proto.

Fixes #558


actien:
Oops sorry, I submitted by accident, didn't do mvn stuff yet

PR #633, opened 2019-09-09
modify grpc-netty dependency as grpc-netty-shaded

riverlee2014:
change the pom.xml dependency of 'jetc-core' module and the java import package, make it dependent on grpc-netty-shaded to reduce netty library exposure.

### DIFF
--- a/jetcd-core/pom.xml
+++ b/jetcd-core/pom.xml
@@ -49,7 +49,7 @@
         </dependency>
         <dependency>
             <groupId>io.grpc</groupId>
-            <artifactId>grpc-netty</artifactId>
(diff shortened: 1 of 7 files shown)


lburgazzoli:
Looks like the PR is not good

lburgazzoli:
@riverlee2014 can we close this as we have now solved the shading issue with jetcd-all ?

riverlee2014:
> @riverlee2014 can we close this as we have now solved the shading issue with jetcd-all ?

Yes, thank you. When will the new version be released?

lburgazzoli:
This week, I hope :)

PR #651, opened 2019-10-30
jetcd-core: overload grant api in Lease interface, add deadline for g…

### LINKED ISSUE #634: LeaseClient grant api caused thread leak
If create client with default threadPool ,  that is `Executors.newCachedThreadPool()` in `ClientConnectionManager`, when using grant api with a ttl for a lease , It make a listenableFuture call in grpc stub, since no deadline or timeout set in, in java it will be a unlimited call util server response。In while, jetcd client make a listenableFuture type to  a
completableFuture  response ,but when to use listenableFuture f.get()，there still no timeout ,so if etcd server cannot response, it will infinite waiting for the listenableFuture call in grpc stub as mentioned before, and the thread for leaseClient.grant will be waiting indifinitely。 since it's a newCachedThreadPool, the threads in JVM will keep rising util the limit of kernel just make s thread leak.

So， why don't set a timeout in listenableFuture f.get() ,just as `targetFuture.complete(resultConvert.apply(f.get(long timeout, TimeUnit unit)))`when converte listenableFuture type to  completableFuture  response for waiting a future response ? 

Best regards

YJFinBJ:
…rpc invocation

Fixes #634

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/LeaseImpl.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/LeaseImpl.java
@@ -106,6 +106,12 @@ final class LeaseImpl implements Lease {
   }
 
   @Override
+  public CompletableFuture<LeaseGrantResponse> grant(long ttl, long timeout, TimeUnit unit) {
+    return connectionManager.execute(() -> this.stub.withDeadlineAfter(timeout,unit).leaseGrant(LeaseGrantRequest
+            .newBuilder().setTTL(ttl).build()),LeaseGrantResponse::new);
+  }
+
+  @Override
   public CompletableFuture<LeaseRevokeResponse> revoke(long leaseId) {
     return connectionManager.execute(
         () -> this.stub.leaseRevoke(LeaseRevokeRequest.newBuilder().setID(leaseId).build()),
--- a/jetcd-core/src/main/java/io/etcd/jetcd/Lease.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/Lease.java
@@ -23,6 +23,7 @@ import io.etcd.jetcd.lease.LeaseTimeToLiveResponse;
 import io.etcd.jetcd.options.LeaseOption;
 import io.grpc.stub.StreamObserver;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Interface of KeepAlive talking to etcd.
@@ -35,6 +36,17 @@ public interface Lease extends CloseableClient {
    * @param ttl ttl value, unit seconds
    */
   CompletableFuture<LeaseGrantResponse> grant(long ttl);
+
+  /**
+   * New a lease with ttl value.Waits if necessary for at most the given time
+   * if etcd server is available.
+   *
+   * @param ttl ttl value, unit seconds
+   * @param timeout the maximum time to waits
+   * @param unit the time unit of the timeout argument
+   * @return
+   */
+  CompletableFuture<LeaseGrantResponse> grant(long ttl, long timeout, TimeUnit unit);
 
   /**
    * revoke one lease and the key bind to this lease will be removed.
--- a/jetcd-core/src/test/java/io/etcd/jetcd/LeaseTest.java
+++ b/jetcd-core/src/test/java/io/etcd/jetcd/LeaseTest.java
@@ -29,7 +29,9 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+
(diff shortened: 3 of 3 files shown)


jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java
```diff
@@ -330,35 +331,84 @@ public void onClose(Status status, Metadata trailers) {
   public <S, T> CompletableFuture<T> execute(
       Callable<ListenableFuture<S>> task,
       Function<S, T> resultConvert) {
-    return execute(task, resultConvert, Util::isRetryable);
+    return execute(task,resultConvert, Util::isRetryable);
```
lburgazzoli: this change is unrelated to the PR

jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java
```diff
@@ -330,35 +331,84 @@ public void onClose(Status status, Metadata trailers) {
...
    * execute the task and retry it in case of failure.
    *
+   * @param task a function that returns a new ListenableFuture.
+   * @param resultConvert a function that converts Type S to Type T.
+   * @param <S> Source type
+   * @param <T> Converted Type.
+   * @return a CompletableFuture with type T.
+   */
+  public <S, T> CompletableFuture<T> execute(
+          Callable<ListenableFuture<S>> task,long timeout,TimeUnit unit,
```
lburgazzoli: would be nice to keep a consistent style

jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java
```diff
@@ -330,35 +331,84 @@ public void onClose(Status status, Metadata trailers) {
...
+   * @param unit the time unit of the timeout argument.
    * @param resultConvert a function that converts Type S to Type T.
    * @param doRetry a function that determines the retry condition base on SourceFuture error.
    * @param <S> Source type
    * @param <T> Converted Type.
    * @return a CompletableFuture with type T.
    */
   public <S, T> CompletableFuture<T> execute(
-      Callable<ListenableFuture<S>> task,
+      Callable<ListenableFuture<S>> task, long timeout, TimeUnit unit,
```
lburgazzoli: would be nice to keep a consistent style

jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java
```diff
@@ -330,35 +331,84 @@ public void onClose(Status status, Metadata trailers) {
...
+
+    return Failsafe.with(retryPolicy).with(executorService)
+            .getAsync(() -> task.call().get())
+            .thenApply(resultConvert);
+  }
+
+  private <S> RetryPolicy<S> getsRetryPolicy(Predicate<Throwable> doRetry) {
+    RetryPolicy<S> retryPolicy = new RetryPolicy<S>().handleIf(doRetry::test).onRetriesExceeded(e -> newEtcdException(
+            ErrorCode.ABORTED, "maximum number of auto retries reached")).withBackoff(builder.retryDelay(), builder
+            .retryMaxDelay(), builder.retryChronoUnit());
```
lburgazzoli: the style here is quite not easy to understand, better to reformat it a little

jetcd-core/src/main/java/io/etcd/jetcd/LeaseImpl.java
```diff
@@ -105,6 +105,12 @@
     );
   }
 
+  @Override
+  public CompletableFuture<LeaseGrantResponse> grant(long ttl, long timeout, TimeUnit unit) {
+    return connectionManager.execute(() -> this.stub.withDeadlineAfter(timeout,unit).leaseGrant(LeaseGrantRequest
+            .newBuilder().setTTL(ttl).build()),timeout,unit,LeaseGrantResponse::new);
```
lburgazzoli: here the same timeout is applied to two different concept which is quite confusing, would be better to use only one of the two

jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java
YJFinBJ: here I just extract  common code, and do not make any logic change

jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java
YJFinBJ: here,it a codestyle conflict via my local workspace. I withdraw this commit

jetcd-core/src/main/java/io/etcd/jetcd/LeaseImpl.java
YJFinBJ: um, I use as  a grpc invocation timeout then

PR #664, opened 2019-11-25
Fix EtcdClusterResource invocation in README

jdhoek:
`EtcdClusterResource#cluster()` does not exist in current releases. This replaces the example code with the direct call to `#getClientEndpoints()`.


xiang90:
thanks

Issue #659, opened 2019-11-06
When Token is invalid the watcher  will be hung there


johnyannj:
jetcd 0.4.1

1、For normal request , like  kv.get

when token is invalid, the flow will enter 

io.grpc.ForwardingClientCallListener.SimpleForwardingClientCallListener#onClose

```
super.start(new SimpleForwardingClientCallListener<RespT>(responseListener) {
            @Override
            public void onClose(Status status, Metadata trailers) {
              if (isInvalidTokenError(status)) {
                try {
                  refreshToken(next);
                } catch (Exception e) {
                  // don't throw any error here.
                  // rpc will retry on expired auth token.
                }
              }
              super.onClose(status, trailers);
            }
          }, headers);
```

2、for watch request

when token is invalid, the flow will never enter "onClose" to  “refreshToken”

it will receive a normal message "Permission Deny", 

io.etcd.jetcd.WatchImpl.WatcherImpl#onNext  process the message and return normal.  it will not try again.

```
 if (response.getWatchId() == -1) {
          listener.onError(newEtcdException(ErrorCode.INTERNAL, "etcd server failed to create watch id"));
          return;
        }
```

There is no chance for watcher to refresh Token,. the watcher is hanging there....

lukasz-antoniak:
I tried to replicate the issue, but token refresh functionality works correctly. `AuthTokenInterceptor` gets triggered when token expires and watch does not hang infinitely. Submitted pull request #728 to include this scenario in regression tests.

lukasz-antoniak:
Figured out the issue. Watcher hangs when token is invalid and we try to register new watch. I will try to fix it and open new PR.

lukasz-antoniak:
New PR should solve this issue - #729.

lburgazzoli:
Fixed by #729

PR #728, opened 2020-03-29
Test that watch refreshes access token correctly

lukasz-antoniak:
Regression test for issue #659 showing that token refresh is performed correctly.


lukasz-antoniak:
Was able to replicate the issue. It happens when token is invalid at the moment we create a watch. Closing this PR, will open a new one after fixing the behavior.

PR #753, opened 2020-05-16
Add keepalive config for gRPC

Excpt0r:
It should be possible to provide the underlying gRPC with keepalive configuration parameters. My use case is to detect during a watch on client side, when the ETCD server is gone silently.

I didn't want to change the default behaviour of jetcd, but it may be worth to think about enabling keepalive per default. K8s seems to use 30s as default (https://github.com/kubernetes/kubernetes/blob/9d3406c27b581c0961ac5871f6893f838d59b10c/staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go#L49)

Cheers,
Alex


lburgazzoli:
@Excpt0r no objection to change the default behavior

Excpt0r:
@lburgazzoli thanks, I updated the default behaviour.

Issue #778, opened 2020-07-08
Slow grpc re-connect


Excpt0r:
Hi,

I'm developing an application that needs a leader/follower mechanism for high availability.
The implementation is done by using jetcd as client for an ETCD cluster, to store the current leader. Followers watch the leader key.

To detect when the ETCD instance of a watching follower is gone, gRPC keepalive is activated.
This works good, but it takes about 12 seconds between the client detecting the gone connection and the re-connect to another ETCD instance. As far as I could figure out in `ClientBuilder.java` the reconnect should be after `retryDelay=500ms`.

I'm using jetcd 0.5.4-SNAPSHOT, the logs are attached
[Reconnect.txt](https://github.com/etcd-io/jetcd/files/4892125/Reconnect.txt)

jcferretti72:
I think what you describe is commonly known as "leader election" in distributed systems lingo.
I think the generally used implementation on etcd involves a fixed key location to identify the leader (stored in the value), associated to a lease with a TTL, and the leader having to renew periodically that lease; if/when the leader goes away etcd will remove the KV.
followers can become leaders by trying to write to the key and only succeeding when it doesn't exist
Sadly, I don't think you can use watchers for this on etcd, and you need to have followers use a polling strategy.
The reason is watchers can go silent and stop being refreshed if an etcd peer is down or becomes partitioned, depending on which peer your client watch is connected to.  Watches are in this sense eventually consistent but not fully consistent; when the etcd cluster heals itself the watcher will continue to refresh, but while the peer is down or partitioned it will not.  This prevents your leader/follower mechanism to be truly fault tolerant since the etcd cluster itself can continue to operate with a peer down or partitioned, but your leader follower mechanism will freeze working/won't update if your leader crashes and a new leader is necessary while one etcd server peer is down, if the followers trying to  become leaders happened to be connected to that peer.
With a polling strategy doing periodic GET (Range) requests with serializable=false (the default, which makes the read fully consistent / linearizable, see here: https://github.com/etcd-io/etcd/blob/master/Documentation/learning/api.md#range)

-Cristian

Excpt0r:
@jcferretti72 thank you, I appreciate your input about the design!

I already experienced the problem that a watching client doesn't detect when the server has gone silently.
The solution I found is to activate keep-alive in the underlying grpc (it's available since jetcd 0.5.4, see here https://github.com/etcd-io/jetcd/pull/753)
[...]

PR #819, opened 2020-09-06
Handling of hasleader metadata annotation and errors related to it.

jcferretti72:
Some previous discussion in https://github.com/etcd-io/jetcd/issues/778#issuecomment-657763381

This diff
(1) Adds special handling for the Status(UNAVAILABLE, "etcdserver: no leader") watcher error.  The added code will prevent retrying/recreating the watcher (previously on calls to WatcherImpl.onError the watch will unconditionally be retried, WatchImpl.java:279).  Instead, the watch is closed, similar to logic in the etcd go client:
     https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/watch.go#L649
(2) Adds a "WithRequireLeader" WatchOption.  This is necessary since automated watch retries can happen outside of a client-initiated context for a WatchClient.watch() call, so it is not possible to use io.grpc.Context to set the metadata annotation reliably.
(3) Adds the hasleader,true metadata annotation to the lease stub of LeaseImpl; note the current go client does this: 
     https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/lease.go#L209
     https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/lease.go#L474
(4) Adds a unit test for the handling of the no leader status.

I also tested the behavior in with 3 machine etcd cluster; the machines were connected to the same switch in a LAN; I had a test client program connected through localhost to one of the etcd peers, as its single etcd endpoint.

The test program creates a watch on a single key and prints updates as they appear.

When created without the require leader option, as I disconnect the machine that runs the client and etcd peer from the switch, the client continues to run and wait for updates that never show up, also never failing.  Effectively the watcher goes silent as the cluster continues to allow updates to the key (as the quorum stays in the two remaining machines).

When created with the require leader option, a few seconds after disconnecting the machine onError is triggered on the listener and the watcher closes.

I am happy to share that code too although I am not sure what a good place for that is, since it is more of an integration test than a system test; creating a docker version of that test is beyond my docker capabilities sadly.

Side comment:  It seems to me the behavior or io.etcd.jetcd.Listener is different from io.grpc.stub.StreamObserver, although there is a good deal of similitude (all methods signatures match).  There seems to be a difference in the expected behavior after onError is called: the grpc StreamObserver contract is to never call any other methods after onError has been called https://grpc.github.io/grpc-java/javadoc/io/grpc/stub/StreamObserver.html#onError-java.lang.Throwable-
Since jetcd.Watcher retries the watcher, in a jetcd.Listener you effectively can get calls to onNext after a call to onError.  Perhaps I am not understanding this well; in any case, I don't think there is any problem in the jetcd.Listener contract, as long as it is explicit and understood as such.  My issue is that the similarity to grpc.stub.StreamObserver is a potential user pitfall; while reviewing uses of jetcd watchers in our own code I found 2 instances of an assumption that the watcher would be closed after onError was called on the listener, and our code was re-creating it against the same listener; I believe this would result in a bug since the listener will be notified twice after that.  The programmer coding this was used to the grpc API so their assumption was understandable.  I humbly suggest javadoc around jetcd.Listener to clarify/make this explicit; granted, the current javadoc does say "invoked on errors" (plural), but still, some additional text making the distinction more explicit might help.  Happy to provide that in a separate pull if you agree.  Using different signatures for the methods would be perhaps better, but that is an API breaking change so has its own pitfalls.

Thanks for your time!

### DIFF
--- a/jetcd-common/src/main/java/io/etcd/jetcd/common/exception/EtcdException.java
+++ b/jetcd-common/src/main/java/io/etcd/jetcd/common/exception/EtcdException.java
@@ -16,6 +16,7 @@
 
 package io.etcd.jetcd.common.exception;
 
+import io.grpc.Status;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
@@ -35,5 +36,12 @@ public class EtcdException extends RuntimeException {
      */
     public ErrorCode getErrorCode() {
         return code;
+    }
+
+    /**
+     * Converts to Status
+     */
+    public Status toStatus() {
+        return Status.fromCodeValue(getErrorCode().getCode()).withDescription(getMessage()).withCause(getCause());
     }
 }
--- a/jetcd-common/src/main/java/io/etcd/jetcd/common/exception/EtcdException.java
+++ b/jetcd-common/src/main/java/io/etcd/jetcd/common/exception/EtcdException.java
@@ -16,7 +16,6 @@
 
 package io.etcd.jetcd.common.exception;
 
-import io.grpc.Status;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
@@ -36,12 +35,5 @@ public class EtcdException extends RuntimeException {
      */
     public ErrorCode getErrorCode() {
         return code;
-    }
-
-    /**
-     * Converts to Status
-     */
-    public Status toStatus() {
-        return Status.fromCodeValue(getErrorCode().getCode()).withDescription(getMessage()).withCause(getCause());
     }
 }
--- a/jetcd-core/src/main/java/io/etcd/jetcd/ClientBuilder.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/ClientBuilder.java
@@ -479,10 +479,11 @@ public final class ClientBuilder implements Cloneable {
 
     /**
      * @param  connectTimeoutMs Sets the connection timeout in milliseconds.
-     * @return                  Clients connecting to fault tolerant etcd clusters (eg, clusters with >= 3 etcd server
+     *                          Clients connecting to fault tolerant etcd clusters (eg, clusters with >= 3 etcd server
      *                          peers/endpoints)
      *                          should consider a value that will allow switching timely from a crashed/partitioned peer to
      *                          a consensus peer.
+     * @return                  this builder
      */
     public ClientBuilder connectTimeoutMs(Integer connectTimeoutMs) {
         this.connectTimeoutMs = connectTimeoutMs;
--- a/jetcd-core/src/main/java/io/etcd/jetcd/Constants.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/Constants.java
@@ -16,6 +16,8 @@
 
 package io.etcd.jetcd;
 
+import io.grpc.Metadata;
+
 /**
  * Constants of Etcd.
  */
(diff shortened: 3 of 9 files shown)


lburgazzoli:
@jcferretti72 the PR looks good and about the side note related to the Listener interface you are right so wonder if we can make the behaviour more in line with the StreamObserver

jcferretti72:
@lburgazzoli I think the go client does retries in the background without communicating errors up to the watch client API, I guess this is similar to what gRPC does for instance when it does transparent reconnection/retries, so you only see an error in an API call when retries failed (most commonly because you as a client set some timeout and that is being hit, the gRPC timeouts tend to be pretty generous, say 30 seconds).

So we could do something similar here, ie, keep exactly the same code and just stop calling listener.onError on non-retry-stopping errors (all before, most with this change).  That change would not be, strictly speaking, a jetcd API breaking change, since we would be sending up less errors than before, but it still would be a pragmatic behavioral change; conceivably some jetcd users could be taking advantage of the info flowing up in terms of errors seen but that seems like a stretch, dunno.

In any case, separate pull request?  Before making any changes to this I think I would like to take a closer look at the go client and ensure I am understanding right how errors are handled there and what is being passed as an error to the API and what isn't.   Is more complicated because I think the go client uses a single stream to multiplex multiple watch requests (difference in Context for the watch call force different streams but other than that is the same stream), which complicates the code a lot but I guess is more efficient.

lburgazzoli:
> @lburgazzoli I think the go client does retries in the background without communicating errors up to the watch client API, I guess this is similar to what gRPC does for instance when it does transparent reconnection/retries, so you only see an error in an API call when retries failed (most commonly because you as a client set some timeout and that is being hit, the gRPC timeouts tend to be pretty generous, say 30 seconds).
> 
> So we could do something similar here, ie, keep exactly the same code and just stop calling listener.onError on non-retry-stopping errors (all before, most with this change). That change would not be, strictly speaking, a jetcd API breaking change, since we would be sending up less errors than before, but it still would be a pragmatic behavioral change; conceivably some jetcd users could be taking advantage of the info flowing up in terms of errors seen but that seems like a stretch, dunno.
> 

It is fine with me however I remember there were people asking to get a notification for each error but eventually that could be made optional or they can disable retry. Or we can distinguish between retryable and non retryable onse by using different methods.

> In any case, separate pull request? Before making any changes to this I think I would like to take a closer look at the go client and ensure I am understanding right how errors are handled there and what is being passed as an error to the API and what isn't. Is more complicated because I think the go client uses a single stream to multiplex multiple watch requests (difference in Context for the watch call force different streams but other than that is the same stream), which complicates the code a lot but I guess is more efficient.

The original code had a single stream here too but I didn't it as it was source for any sort of errors and misbehaviour. My plan was to support a single stream when moving to etcd 3.4 that has an option to set a request id instead of rely on the one generated by the server.

lburgazzoli:
@jcferretti72 I'd like to merge this PR but looks like there's some failures, mind having a look ?

jcferretti72:
Sorry I missed that.  There was an issue with the handling of permission denied and the check on whether to reschedule the watcher or not.  Fix pushed, PTAL.

lburgazzoli:
@jcferretti72 thank you for the PR

lburgazzoli:
@jcferretti72 mind opening an issue for what discussed in https://github.com/etcd-io/jetcd/pull/819#issuecomment-689036406 ?

PR #859, opened 2020-11-30
Remove @Beta annotation to resolve #857

### LINKED ISSUE #857: question: Beta annotation for KV and Election clients
Hello,

There are 2 clients, KV and Election, which are annotated with `@Beta` meaning that "is declared as unstable interface"

May I ask why is still unstable and if there is a plan to make it stable?

Thanks for your time on this!

/question

marcosflobo:
Remove @Beta annotations which were leftovers

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/Election.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/Election.java
@@ -18,7 +18,6 @@ package io.etcd.jetcd;
 
 import java.util.concurrent.CompletableFuture;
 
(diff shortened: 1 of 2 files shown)


lburgazzoli:
@marcosflobo can you check why the build is failing ?

marcosflobo:
@lburgazzoli it looks like the pipeline just took too long https://github.com/etcd-io/jetcd/actions/runs/392157377
```
The job running on runner Hosted Agent has exceeded the maximum execution time of 360 minutes.
```
Is there a way to re-run the pipeline without a new commit?

Issue #844, opened 2020-11-02
Unable to use TLS since 0.5.4


bh-tt:
We have an etcd cluster reachable using HTTPS. Since 0.5.4 we were unable to do this due to a TLS error. After some searching we found out jetcd no longer uses `https` schemes in its urls, instead opting for `ip` or `dns+srv`. This results in a target such as `ip:///etcd0.com:2379,etcd1.com:2379,etcd2.com:2379`. We have verified that the cluster is reachable using etcdctl. 

 - Should we simply disable TLS because jetcd uses a different scheme than HTTP?
- If not, how do we configure our certificates to work with the `ip` scheme? I cannot find anything online about this scheme.

## Reproduce
Try to connect to any etcd instance with the following SSLContext, using any common URI such as `https://etcd0.com:2379`:
```
import io.netty.handler.ssl.ApplicationProtocolConfig;
import io.netty.handler.ssl.SslContextBuilder;
                ApplicationProtocolConfig apn = new ApplicationProtocolConfig(
                        ApplicationProtocolConfig.Protocol.ALPN,
                        ApplicationProtocolConfig.SelectorFailureBehavior.NO_ADVERTISE,
                        ApplicationProtocolConfig.SelectedListenerFailureBehavior.ACCEPT,
                        "h2");
                builder = builder
                        .sslContext(SslContextBuilder
                                .forClient()
                                .applicationProtocolConfig(apn)
                                .build());
```

## Expected outcome
The jetcd client verifies the certificate correctly and connects to the etcd client

## Current outcome
```
Caused by: javax.net.ssl.SSLHandshakeException: No subject alternative DNS name matching ip found.
        at java.base/sun.security.ssl.Alert.createSSLException(Alert.java:131)
        at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:326)
        at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:269)
        at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:264)
        at java.base/sun.security.ssl.CertificateMessage$T12CertificateConsumer.checkServerCerts(CertificateMessage.java:645)
        at java.base/sun.security.ssl.CertificateMessage$T12CertificateConsumer.onCertificate(CertificateMessage.java:464)
        at java.base/sun.security.ssl.CertificateMessage$T12CertificateConsumer.consume(CertificateMessage.java:360)
[...]

PR #862, opened 2020-12-07
fix sslcontext not work bug since 0.5.4

lichaojacobs:
Refrenced by the issue:  [https://github.com/etcd-io/jetcd/issues/844](https://github.com/etcd-io/jetcd/issues/844)

we have aslo encountered the problem that we can not use tls through `SslContext` after we upgrate to 0.5.4.  

After a deep dive, I finally found that addresses in  `IPNameResolver` should have default authority when we have not use the manner of passing through authority.

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/resolver/IPNameResolver.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/resolver/IPNameResolver.java
@@ -27,9 +27,11 @@ import javax.annotation.concurrent.GuardedBy;
 
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
+import com.google.common.base.Strings;
 import com.google.common.collect.Iterables;
 import io.etcd.jetcd.common.exception.ErrorCode;
 import io.etcd.jetcd.common.exception.EtcdExceptionFactory;
+import io.grpc.Attributes;
 import io.grpc.EquivalentAddressGroup;
 import io.grpc.NameResolver;
 import io.grpc.Status;
@@ -60,7 +62,7 @@ public class IPNameResolver extends NameResolver {
     public IPNameResolver(URI targetUri) {
         this.lock = new Object();
         this.targetUri = targetUri;
-        this.authority = targetUri.getAuthority() != null ? targetUri.getAuthority() : SCHEME;
+        this.authority = targetUri.getAuthority() != null ? targetUri.getAuthority() : "";
         this.addresses = Stream.of(targetUri.getPath().split(","))
             .map(address -> {
                 return address.startsWith("/")
@@ -71,12 +73,18 @@ public class IPNameResolver extends NameResolver {
                 Iterable<String> split = Splitter.on(':').split(address);
                 String host = Iterables.get(split, 0);
                 String port = Iterables.get(split, 1, ETCD_CLIENT_PORT);
-                return new InetSocketAddress(host, Integer.parseInt(port));
-            }).map(address -> {
                 return new EquivalentAddressGroup(
-                    address,
-                    io.grpc.Attributes.EMPTY);
+                    new InetSocketAddress(host, Integer.parseInt(port)),
+                    Strings.isNullOrEmpty(authority)
+                        ? Attributes.newBuilder()
+                            .set(EquivalentAddressGroup.ATTR_AUTHORITY_OVERRIDE, getDefaultAuthority(host, port))
+                            .build()
+                        : Attributes.EMPTY);
             }).collect(Collectors.toList());
+    }
+
+    private static String getDefaultAuthority(String host, String port) {
+        return String.format("%s:%s", host, port);
     }
 
     @Override


jetcd-core/src/main/java/io/etcd/jetcd/resolver/IPNameResolver.java
```diff
@@ -61,21 +66,33 @@ public IPNameResolver(URI targetUri) {
...
-            .map(address -> {
+
+        List<String> defaultAuthorities = Arrays.stream(this.authority.split(",")).collect(Collectors.toList());
+        List<String> addressStrList = Stream.of(targetUri.getPath().split(","))
+            .map(
+                address -> {
+                    return address.startsWith("/") ? address.substring(1) : address;
+                })
+            .collect(Collectors.toList());
+        final boolean needOverrideAuthority = addressStrList.size() == defaultAuthorities.size();
```
lburgazzoli, line 77: there should be some check here as id the size is not the same, it could be because of a misconfiguration

jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java
```diff
@@ -224,7 +224,8 @@ void close() {
             target = String.format(
                 "%s://%s/%s",
                 IPNameResolver.SCHEME,
-                builder.authority() != null ? builder.authority() : "",
+                builder.authority() != null ? builder.authority()
+                    : endpoints.stream().map(URI::getAuthority).collect(Collectors.joining(",")),
```
lburgazzoli, line 228: URI::getAuthority may be null

lichaojacobs: fixed

jetcd-core/src/main/java/io/etcd/jetcd/resolver/IPNameResolver.java
```diff
@@ -61,21 +66,37 @@ public IPNameResolver(URI targetUri) {
...
         this.authority = targetUri.getAuthority() != null ? targetUri.getAuthority() : SCHEME;
-        this.addresses = Stream.of(targetUri.getPath().split(","))
-            .map(address -> {
-                return address.startsWith("/")
-                    ? address.substring(1)
-                    : address;
-            })
-            .map(address -> {
+
+        List<String> defaultAuthorities = Arrays.stream(this.authority.split(",")).collect(Collectors.toList());
```
lburgazzoli, line 70: don't think is is needed to have a list here , using a String[] should be enough

jetcd-core/src/main/java/io/etcd/jetcd/resolver/IPNameResolver.java
```diff
@@ -61,21 +66,37 @@ public IPNameResolver(URI targetUri) {
...
+            .map(
+                address -> {
+                    return address.startsWith("/") ? address.substring(1) : address;
+                })
+            .collect(Collectors.toList());
+        if (defaultAuthorities.size() > 1) {
+            Preconditions.checkState(addressStrList.size() == defaultAuthorities.size(),
+                "misConfigured endpoint uris");
+        }
+        final boolean needOverrideAuthority = addressStrList.size() == defaultAuthorities.size();
```
lburgazzoli, line 81: this flag can be set in the previous condition

jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java
```diff
@@ -224,8 +225,14 @@ void close() {
             target = String.format(
                 "%s://%s/%s",
                 IPNameResolver.SCHEME,
-                builder.authority() != null ? builder.authority() : "",
-                endpoints.stream().map(e -> e.getHost() + ":" + e.getPort()).collect(Collectors.joining(",")));
+                builder.authority() != null ? builder.authority()
```
lburgazzoli, line 228: I wonder if this is needed as if the authority is empty, then it can be computed automatically out of the URI in the IP resolver

lichaojacobs: ok，i will make some changes to compute authority in Ip resolver

lichaojacobs: I just change `e.getHost() + ":" + e.getPort()` to `URI::toString`, and then in Ip resolver we change re compute URI again.

jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java
```diff
@@ -225,7 +225,7 @@ void close() {
                 "%s://%s/%s",
                 IPNameResolver.SCHEME,
                 builder.authority() != null ? builder.authority() : "",
-                endpoints.stream().map(e -> e.getHost() + ":" + e.getPort()).collect(Collectors.joining(",")));
+                endpoints.stream().map(URI::toString).collect(Collectors.joining(",")));
```
lburgazzoli, line 228: Not very sure it is needed: the authority is defined as`[userinfo@]host[:port]` but given that the `userinfo` do not make any sense here, why do we need toString ? toString also included the scheme which does not make any sense in the ip resolver

jetcd-core/src/main/java/io/etcd/jetcd/resolver/IPNameResolver.java
```diff
@@ -61,21 +60,24 @@ public IPNameResolver(URI targetUri) {
...
-                Iterable<String> split = Splitter.on(':').split(address);
-                String host = Iterables.get(split, 0);
-                String port = Iterables.get(split, 1, ETCD_CLIENT_PORT);
-                return new InetSocketAddress(host, Integer.parseInt(port));
-            }).map(address -> {
                 return new EquivalentAddressGroup(
-                    address,
-                    io.grpc.Attributes.EMPTY);
+                    new InetSocketAddress(address.getHost(), address.getPort()),
+                    needOverrideAuthority
```
lburgazzoli, line 75: this can be simplified by changing the way` this.authority` is handled as we can leave it null if the authority needs to be overridden (of course the getAuthority method need to be updated)

jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java
lichaojacobs: Use `URI::toString` then we can  use `URI::create` in another place. it is like serialize and deserialize, we just use URI as a basic tool to reduce hard code

lburgazzoli: yes but the IP resolver is not expecting a full URI, just host and port

lichaojacobs: all right, we can aslo use host and port to create URI just like  `URI.create(String.format("ip://%s", address))`

lburgazzoli (approved):

Issue #934, opened 2021-06-17
Change prefix param type to boolean.


horizonzy:
the param prefix at get api is ByteString, it's weird, it can be replace by boolean.

https://github.com/etcd-io/jetcd/blob/76637dc28dcf1beef89c5e07cb3a32cbff86231f/jetcd-core/src/main/java/io/etcd/jetcd/options/GetOption.java#L180-L185

horizonzy:
At go version, prefix is a flag which to get value by input key prefix. We just need `set prefix = true or false`, instead of `set prefix = key` again. It makes me confuse, after see the source code, I understant it. 
So make the prefix param easier to understand, modify the param type, user didn't set key as the prefix value again.

lburgazzoli:
Ok, I think we need two thing here otherwise is even more confusing:
- withPrefix does not sound right for a boolean value, it seems to suggest you have to provide a prefix, maybe isPrefix could be better
- it need to be handled consistently in any method not only Get

horizonzy:
That's fine.

horizonzy:
All done, see the newest commit.

lburgazzoli:
Fixed by https://github.com/etcd-io/jetcd/pull/935

PR #935, opened 2021-06-17
Change prefix param type to boolean.

horizonzy:
for-#934.
When user didn't set withRange and prefix is true, use the key which from invocation param to build rangeEnd


horizonzy:
@lburgazzoli can you help to trigger it.

horizonzy:
@lburgazzoli Hi, do me a favor to trigger it, thanks.

lburgazzoli:
@horizonzy anything more you want to include ?

horizonzy:
That's enough.

PR #939, opened 2021-06-21
Override ByteSequence Object.toString().

horizonzy:
image

Some user may use Sequence.toString directly, and didn't notice the toString(Charset charset), So override Object.toString()
by default charset.

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/ByteSequence.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/ByteSequence.java
@@ -136,7 +136,7 @@ public final class ByteSequence {
 
     @Override
     public String toString() {
-        return toString(Charset.defaultCharset());
+        return byteString.toStringUtf8();
     }
 
     /**
--- a/jetcd-core/src/main/java/io/etcd/jetcd/ByteSequence.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/ByteSequence.java
@@ -134,6 +134,11 @@ public final class ByteSequence {
         return byteString.size();
     }
 
+    @Override
+    public String toString() {
+        return toString(Charset.defaultCharset());
+    }
+
     /**
      * Create new ByteSequence from a String.
      *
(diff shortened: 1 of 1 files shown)


lburgazzoli:
I think the problem with this is that you may get different results depending if the one that puts data to etcd and the one reading it have different set-up

horizonzy:
> I think the problem with this is that you may get different results depending if the one that puts data to etcd and the one reading it have different set-up

I'm agree with you. If let user toString with param charset, user will be attention the encoding. The origin api toString(Charset) is deceptive to me. 😂

lburgazzoli:
Unfortunately there's no way to hide `toString()` so I don't know what is the best option here. 
I was not super happy whit this at the time it has been introduced, maybe we need to rename it to `asString` and document that to string is not meant to return the content ...

don't know

horizonzy:
I notifce the ByteString.toString, it show the content in particular way. 
```
  @Override
  public final String toString() {
    return String.format(
        Locale.ROOT,
        "<ByteString@%s size=%d contents=\"%s\">",
        toHexString(identityHashCode(this)),
        size(),
        truncateAndEscapeForDisplay());
  }
```
Now the toString retuen Object.toString(hashCode), it's not a good operation. Maybe Override the Object.toString by ByteString.toUtf8String is a not bad way. UTF_8 encoding is the general encoding way at go. If user setted other encoding, make user find the root case and use toString(Charset). 
And we can add doc to attension user to notice it.(although the user didn't notice it most of the time)

lburgazzoli:
can you rephrase ?

horizonzy:
> can you rephrase ?

I mean that we supply the general encoding utf_8 as the default toString. For me, utf_8 is my first choice most of the time. and golang use utf_8 as the default encoding.  
So we use utf_8 encoding in toString. If some user use the different encoding at server side, and use toString directly, it need them to find out the encoding problem. We just supply general way. And we can add doc to notice user use toString(Charset) to use particular encoding.

PR #950, opened 2021-07-13
Close cancelled watchers

### LINKED ISSUE #951: Cancelled watchers should be closed
When jetcd receives a cancelled response, such as when trying to read a compacted revision and getting a `CompactedException`, the watcher is effectively dead and will not receive any more events. I think it makes sense for jetcd to automatically close the watcher rather than leaving it to the user.

mfleming:
A cancelled watcher is effectively dead and will not receive any more
events so it's helpful for users if the watcher gets closed
automatically.

Note that we don't want to reschedule the watcher since users may want
to take corrective action first, such as finding the latest revision if
the watcher was cancelled due to CompactedException.

Closes #951


jtgrabowski:
@lburgazzoli could you please take a look?

Issue #970, opened 2021-08-31
Detecting connection errors


carlanton:
Hi!

What is the correct way to do KV requests with a timeout? The client seems to retry forever (jetcd 0.5.10).
My code looks like this:

```java
Client etcd = Client.builder()
        .endpoints("http://127.0.0.1:2379")
        .connectTimeout(Duration.ofSeconds(1))
        .build();

CountDownLatch countDownLatch = new CountDownLatch(1);
etcd.getKVClient()
        .get(ByteSequence.from("/x", StandardCharsets.UTF_8))
        .whenComplete((response, throwable) -> {
            if (throwable != null) {
                throwable.printStackTrace();
            }

            System.out.println("response = " + response);
            countDownLatch.countDown();
        });

countDownLatch.await();
System.out.println("done");
```

If I run this code without ectd running, it will never terminate. I've tried different retry options to the client builder but I can't get it to work.

Thanks,
Anton

lburgazzoli:
you mean, you run the code without having etcd running at all ?
can you please create a test/reproducer for this issue ?

carlanton:
Exactly! And I want to get notified if my application fails to connect the etcd server. I'll try to create a test.

carlanton:
Hopefully this test (#971) makes sense.

Currently, when I do `future.get(10, TimeUnit.SECONDS)` I will get a TimeoutException since the future is not ready yet. The thing I would except is that the future completes exceptionally (because of too-many-reties or something), which then can be handeld with the CompletableFuture api. What do you think?

lburgazzoli:
thx for the PR, I'll try to have a look at the issue in the coming days but I can't promise any ETA as I'm quite busy at the moment so if you have any spare time to digg into the issue, then it would be really appreciated.

carlanton:
I think that the problem is related to gRPC's "[wait for ready](https://github.com/grpc/grpc/blob/master/doc/wait-for-ready.md)" semantics. Currently it is enabled for all stubs (`.withWaitForReady()` in ClientConnectionManager). I updated the PR with a new setting in ClientBuilder that can disable it.

Another option could be to disable wait-for-retry completely  and rely on the Failsafe retry thing that is used for other kinds of failures...

lburgazzoli:
[...]

PR #971, opened 2021-08-31
test: kv requests should give up at some point

carlanton:
test that should detect #970


lburgazzoli:
@carlanton the Ci seems to have skipped the tests, mind force pushing something so we can see if the action can get re-triggered ?

carlanton:
Sorry for the delay. I think you need to approve me or something... https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

carlanton:
My local build setup is a bit wonky, but now the tests should work ... Sorry.

PR #987, opened 2021-09-24
The grpc request to get the token is also added with a custom header

thislyc:


lburgazzoli (changes requested):
Please provide a unit test

thislyc:
I find it difficult to write a unit test of this function. The interceptor cannot intercept the grpc request to obtain the token in the test class

thislyc:
In my own project, I adopted this modification method and successfully transmitted the header. But I can't show it in the test class.

lburgazzoli:
> In my own project, I adopted this modification method and successfully transmitted the header. But I can't show it in the test class.

I understand but we need a test. 
I really have very little time to maintain the project so I need feature to be tested.

thislyc:
> > Okay, this class [`3d88fde`/jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java](https://github.com/etcd-io/jetcd/blob/3d88fded7fb022c627ed9495c5cff1d4f282630d/jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java)
> 
> Yes I know :) What I meant is that you can probably mock the grp server and inspect the headers that have been sent, something similar to what https://github.com/etcd-io/jetcd/blob/3d88fded7fb022c627ed9495c5cff1d4f282630d/jetcd-core/src/test/java/io/etcd/jetcd/LeaseUnitTest.java does

Hello, I've checked this test class carefully. However, it can't solve my problem. The request to obtain the token is sent in the authtokeninterceptor of clientconnectionmanager.java. I can't add an interceptor to this request. Please look at the code I modified. There is only a little logic. You can easily understand what I mean. It only takes a few minutes. Trust me.
I've spent several days thinking about how to write this test class, but it's really special. I can't get inspiration from other test classes. I'm about to give up this pr😭

lburgazzoli:
I'll try to have a look in the coming days

PR #1023, opened 2021-12-02
fix gradle dependencies in README

daleiz:
In Gradle 7, both the ``compile`` and ``runtime`` configurations are removed.
see, https://docs.gradle.org/current/userguide/upgrading_version_6.html#sec:configuration_removal


lburgazzoli (approved):

lburgazzoli:
thx

PR #1024, opened 2021-12-03
feat: switch to vert.x as grpc impl

lburgazzoli:


sudoHackIn:
Hi, @lburgazzoli.
Sorry, for my curiosity,  but why do you decided to replace netty with vertx?

sudoHackIn:
Thank you, for the explanation

PR #1026, opened 2021-12-08
fix: key prefix support int transactions (Put and GetOpt)

### LINKED ISSUE #1017: isPrefix in GetOp don't work
**Describe the bug**
jetcd version 0.5.10
I tryed to get range values by prefix in txn. But I can't get any value by using isPrex(). When I replace "isPrefix()" by "withPrefix()", it works.

**To Reproduce**
```
void test() throws ExecutionException, InterruptedException {
        Client client = Client.builder().endpoints("http://localhost:2379").build();
        Lease lease = client.getLeaseClient();
        String lockName = "/myLock";

        Long leaseId = lease.grant(10).get().getID();

        KV kvClient = client.getKVClient();
        String lockKey = lockName + "/" + "owner";
        Cmp cmp = new Cmp(ByteSequence.from(lockKey.getBytes()), Cmp.Op.EQUAL, CmpTarget.createRevision(0));
        Op.PutOp putOp = Op.put(ByteSequence.from(lockKey.getBytes()), ByteSequence.from("0".getBytes()), PutOption.newBuilder().withLeaseId(leaseId).build());
        Op.GetOp getOwner = Op.get(ByteSequence.from(lockName, Charset.defaultCharset()), GetOption.newBuilder().isPrefix(true).build());

        TxnResponse response = kvClient.txn().If(cmp).Then(putOp, getOwner).Else(getOwner).commit().get();

        for (PutResponse putResponse : response.getPutResponses()) {
            log.info("transaction putResponse {}", putResponse);
        }

        for (GetResponse getResponse : response.getGetResponses()) {
            log.info("transaction getResponse {}", getResponse);
        }

        GetResponse getResponse = kvClient.get(ByteSequence.from(lockName, Charset.defaultCharset()), GetOption.newBuilder().isPrefix(true).build()).get();
        log.info("getResponse {}", getResponse);
    }
```

In the response there is nothing printed. But In getResponse the value is printed.

And when I replace isPrefix by withPrefix, it works.
```
void test2() throws ExecutionException, InterruptedException {
        Client client = Client.builder().endpoints("http://localhost:2379").build();
        Lease lease = client.getLeaseClient();
        String lockName = "/myLock";

        Long leaseId = lease.grant(10).get().getID();

        KV kvClient = client.getKVClient();
        String lockKey = lockName + "/" + "owner";
        Cmp cmp = new Cmp(ByteSequence.from(lockKey.getBytes()), Cmp.Op.EQUAL, CmpTarget.createRevision(0));
        Op.PutOp putOp = Op.put(ByteSequence.from(lockKey.getBytes()), ByteSequence.from("0".getBytes()), PutOption.newBuilder().withLeaseId(leaseId).build());
        Op.GetOp getOwner = Op.get(ByteSequence.from(lockName, Charset.defaultCharset()), GetOption.newBuilder().withPrefix(ByteSequence.from(lockName, Charset.defaultCharset())).build());

        TxnResponse response = kvClient.txn().If(cmp).Then(putOp, getOwner).Else(getOwner).commit().get();

        for (PutResponse putResponse : response.getPutResponses()) {
            log.info("transaction putResponse {}", putResponse);
        }

        for (GetResponse getResponse : response.getGetResponses()) {
            log.info("transaction getResponse {}", getResponse);
        }

        GetResponse getResponse = kvClient.get(ByteSequence.from(lockName, Charset.defaultCharset()), GetOption.newBuilder().isPrefix(true).build()).get();
        log.info("getResponse {}", getResponse);
    }
```
        

**Expected behavior**
isPrefix works as it works in get method

**Additional context**
I found this in get method in KVImpl.java 
https://github.com/etcd-io/jetcd/blob/release-0.5.10/jetcd-core/src/main/java/io/etcd/jetcd/KVImpl.java#L110
```
       if (!option.getEndKey().isPresent() && option.isPrefix()) {
            ByteSequence endKey = OptionsUtil.prefixEndOf(key);
            builder.setRangeEnd(Util.prefixNamespaceToRangeEnd(endKey.getByteString(), namespace));
        }
```
But same code in GetOp class in Op.java is not existed.
https://github.com/etcd-io/jetcd/blob/release-0.5.10/jetcd-core/src/main/java/io/etcd/jetcd/op/Op.java#L108

sudoHackIn:
Because grpc request mapping is the same for KV and Transactions, it's better for me to reuse it.
So the simplest solution was - additional static util class with this logic.

After of package changing,  [ByteSequence.toByteString](https://github.com/etcd-io/jetcd/blob/0537b7fe62f8985d2a0de554d52d2c05ec90789a/jetcd-core/src/main/java/io/etcd/jetcd/ByteSequence.java#L117) is not longer possible. So, i have additional conversion's, which i don't like. 

Fixes #1017

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/op/Op.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/op/Op.java
@@ -17,10 +17,7 @@
 package io.etcd.jetcd.op;
 
 import io.etcd.jetcd.ByteSequence;
-import io.etcd.jetcd.Util;
-import io.etcd.jetcd.api.DeleteRangeRequest;
-import io.etcd.jetcd.api.PutRequest;
-import io.etcd.jetcd.api.RangeRequest;
+import io.etcd.jetcd.ProtoRequestMapper;
 import io.etcd.jetcd.api.RequestOp;
 import io.etcd.jetcd.api.TxnRequest;
 import io.etcd.jetcd.options.DeleteOption;
@@ -28,9 +25,6 @@ import io.etcd.jetcd.options.GetOption;
 import io.etcd.jetcd.options.PutOption;
 
 import com.google.protobuf.ByteString;
-
-import static io.etcd.jetcd.options.OptionsUtil.toRangeRequestSortOrder;
-import static io.etcd.jetcd.options.OptionsUtil.toRangeRequestSortTarget;
 
 /**
  * Etcd Operation.
@@ -83,10 +77,10 @@ public abstract class Op {
 
         @Override
         RequestOp toRequestOp(ByteSequence namespace) {
-            PutRequest put = PutRequest.newBuilder().setKey(Util.prefixNamespace(this.key, namespace)).setValue(this.value)
-                .setLease(this.option.getLeaseId()).setPrevKv(this.option.getPrevKV()).build();
-
-            return RequestOp.newBuilder().setRequestPut(put).build();
+            return RequestOp.newBuilder()
+                .setRequestPut(
+                    ProtoRequestMapper.mapPutRequest(ByteSequence.from(key), ByteSequence.from(value), option, namespace))
+                .build();
         }
     }
 
@@ -101,18 +95,9 @@ public abstract class Op {
 
         @Override
         RequestOp toRequestOp(ByteSequence namespace) {
-            RangeRequest.Builder range = RangeRequest.newBuilder().setKey(Util.prefixNamespace(this.key, namespace))
-                .setCountOnly(this.option.isCountOnly()).setLimit(this.option.getLimit())
-                .setRevision(this.option.getRevision()).setKeysOnly(this.option.isKeysOnly())
-                .setSerializable(this.option.isSerializable())
-                .setSortOrder(toRangeRequestSortOrder(this.option.getSortOrder()))
-                .setSortTarget(toRangeRequestSortTarget(this.option.getSortField()));
-
-            this.option.getEndKey()
-                .map(endKey -> Util.prefixNamespaceToRangeEnd(ByteString.copyFrom(endKey.getBytes()), namespace))
-                .ifPresent(range::setRangeEnd);
-
-            return RequestOp.newBuilder().setRequestRange(range).build();
+            return RequestOp.newBuilder()
+                .setRequestRange(ProtoRequestMapper.mapRangeRequest(ByteSequence.from(key), option, namespace))
+                .build();
         }
     }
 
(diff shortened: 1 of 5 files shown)


jetcd-core/src/main/java/io/etcd/jetcd/op/Op.java
```diff
@@ -29,8 +25,7 @@
 
 import com.google.protobuf.ByteString;
 
-import static io.etcd.jetcd.options.OptionsUtil.toRangeRequestSortOrder;
-import static io.etcd.jetcd.options.OptionsUtil.toRangeRequestSortTarget;
+import static io.etcd.jetcd.support.ProtoRequestMapper.*;
```
lburgazzoli, line 28: please don't use wildcard import as it does not make it clear what is being imported

jetcd-core/src/main/java/io/etcd/jetcd/support/ProtoRequestMapper.java
```diff
@@ -0,0 +1,73 @@
...
+import io.etcd.jetcd.options.GetOption;
+import io.etcd.jetcd.options.OptionsUtil;
+import io.etcd.jetcd.options.PutOption;
+
+import com.google.protobuf.ByteString;
+
+import static io.etcd.jetcd.options.OptionsUtil.toRangeRequestSortOrder;
+import static io.etcd.jetcd.options.OptionsUtil.toRangeRequestSortTarget;
+
+public class ProtoRequestMapper {
```
lburgazzoli, line 21: this can be made final

lburgazzoli: Also, it can be moved to the same package as the `Util` class

lburgazzoli (changes requested):
Some small things to be fixed and some tests are failing

jetcd-core/src/main/java/io/etcd/jetcd/op/Op.java
sudoHackIn: ok, i see

jetcd-core/src/main/java/io/etcd/jetcd/support/ProtoRequestMapper.java
sudoHackIn: the main one - io.etcd.jetcd? Ok, it'll help with additional boxing and  ByteSequence.toByteString (that is package private)

lburgazzoli: yes, we may think to have better api vs impl separation in the future like when we can use sealed classes/interfaces

sudoHackIn:
> Some small things to be fixed and some tests are failing

fixed

sudoHackIn:
How do you think, do we still need support java 1.8 java as sometimes requested? If yes, i can try to implement support for this

lburgazzoli:
> How do you think, do we still need java 1.8 java support, as sometimes requested? If yes, i can try to implement support for java, 1.8

if you have time, that would be ok however I'd really like to drop Java 1.8

lburgazzoli (approved):

PR #1035, opened 2021-12-15
feat: java 8 support

sudoHackIn:


sudoHackIn:
i thought, that can verify correct build with ci tool, but unfortunately, no execution

lburgazzoli:
@sudoHackIn there are conflicts

gradle.properties
```diff
@@ -14,7 +14,7 @@ protocVersion = 3.17.2
 # plugins
 #
 versionsPluginVersion = 0.39.0
-versionsFilterPluginVersion = 0.1.10
+versionsFilterPluginVersion = 0.1.4
```
lburgazzoli, line 17: this seems to be problematic to me, like get stuck with an older version of something and not getting fixed updates

jetcd-core/src/test/java/io/etcd/jetcd/MaintenanceTest.java
```diff
@@ -92,7 +92,7 @@ public void testSnapshotChunks() throws ExecutionException, InterruptedException
         final AtomicLong count = new AtomicLong();
         final CountDownLatch latcht = new CountDownLatch(1);
 
-        maintenance.snapshot(new StreamObserver<>() {
+        maintenance.snapshot(new StreamObserver<SnapshotResponse>() {
```
lburgazzoli, line 95: is this really needed for Java 8

jetcd-core/src/test/java/io/etcd/jetcd/WatchTokenExpireTest.java
```diff
@@ -45,9 +47,10 @@
     public static final EtcdClusterExtension cluster = EtcdClusterExtension.builder()
         .withNodes(1)
         .withSsl(true)
-        .withAdditionalArgs(List.of(
+        .withAdditionalArgs(Stream.of(
```
lburgazzoli, line 50: Arrays.asList() may be simpler

jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/Etcd.java
```diff
@@ -49,12 +48,12 @@ public Builder withSsl(boolean ssl) {
...
 
         public Builder withAdditionalArgs(Collection<String> additionalArgs) {
-            this.additionalArgs = List.copyOf(additionalArgs);
+            this.additionalArgs = Collections.unmodifiableList(new ArrayList<>(additionalArgs));
             return this;
         }
 
         public Builder withAdditionalArgs(String... additionalArgs) {
-            this.additionalArgs = List.of(additionalArgs);
+            this.additionalArgs = Collections.unmodifiableList(Arrays.stream(additionalArgs).collect(Collectors.toList()));
```
lburgazzoli, line 56: Arrays.asList

obigcat:
@lburgazzoli I really also require java 8 support, I recently completed same things like that. I saw a lot of progress in this project, what's your plan for it ?

lpcy:
@lburgazzoli I really also require java 8 support, I recently completed same things like that. I saw a lot of progress in this project, what's your plan for it ?

lburgazzoli:
@lpcy 

I really recommend to switch your project to Java 11 :)
I can bring back java 8 support but don't expect to last long
Bear in mind that I'm the only maintainer at the moment so I'm trying to reduce the surface of what I have to support to the minimum.

gradle.properties
sudoHackIn: yeah, i understand. But it's the only simple way. Otherwise we should find other plugin, or implement ourselves

sudoHackIn:
> @lburgazzoli I really also require java 8 support, I recently completed same things like that. I saw a lot of progress in this project, what's your plan for it ?

I did this pr only for those, who ask for java 8 support, but soon java 17 will be new LTS, so java 8 won't be long and only if @lburgazzoli decides to include java8 support,

jetcd-core/src/test/java/io/etcd/jetcd/MaintenanceTest.java
sudoHackIn: i'll double check, but looks like, that compiler can't infer this type and asks for 
`public void onNext(Object value)` otherwise

lpcy:
> @lpcy
> 
> I really recommend to switch your project to Java 11 :) I can bring back java 8 support but don't expect to last long Bear in mind that I'm the only maintainer at the moment so I'm trying to reduce the surface of what I have to support to the minimum.

It's hard to switch all of my projects to java 11. Despite the version is old, the proportion of using java 8 is still very high. I can understand your work that switch to java 11 due to easier development. But I suggest you support java 8 until one stable version released such as 1.0, because this project may used for millions developer and be their only best choice.

gradle.properties
lburgazzoli: I understand too but bear in mind that there's no many developer working on this project so I really need to keep the maintenance cost as low as possible. In this particular case I'm fine to provide a custom implementation as it should be very simple but in general, I don't have the time to try to keep up with all those details.

If anyone wants to step in, I'm more than happy

lburgazzoli:
> > @lpcy
> > I really recommend to switch your project to Java 11 :) I can bring back java 8 support but don't expect to last long Bear in mind that I'm the only maintainer at the moment so I'm trying to reduce the surface of what I have to support to the minimum.
> 
> It's hard to switch all of my projects to java 11. Despite the version is old, the proportion of using java 8 is still very high. I can understand your work that switch to java 11 due to easier development. But I suggest you support java 8 until one stable version released such as 1.0, because this project may used for millions developer and be their only best choice.

What I want to do is to support the latest LTS -1, so since java 17 is out and it is LTS then Java 11 should be the minimum requirement. I can bring Java 8 support but if as you said, this project may used for millions developers, we need a fraction of them to contribute back, If that happens, I have no objection to keep Java 8 support for a while but it must be clear that it can't last forever and we need the industry to move forward. 

Likely, 1.0 will be require Java 11 at minimum.

lpcy:
> > > @lpcy
> > > I really recommend to switch your project to Java 11 :) I can bring back java 8 support but don't expect to last long Bear in mind that I'm the only maintainer at the moment so I'm trying to reduce the surface of what I have to support to the minimum.
> > 
> > 
> > It's hard to switch all of my projects to java 11. Despite the version is old, the proportion of using java 8 is still very high. I can understand your work that switch to java 11 due to easier development. But I suggest you support java 8 until one stable version released such as 1.0, because this project may used for millions developer and be their only best choice.
> 
> What I want to do is to support the latest LTS -1, so since java 17 is out and it is LTS then Java 11 should be the minimum requirement. I can bring Java 8 support but if as you said, this project may used for millions developers, we need a fraction of them to contribute back, If that happens, I have no objection to keep Java 8 support for a while but it must be clear that it can't last forever and we need the industry to move forward.
> 
> Likely, 1.0 will be require Java 11 at minimum.

Well, thank you for your reply. I agree with that opinion to move forward, but please consider our request at the moment.

lburgazzoli:
> > What I want to do is to support the latest LTS -1, so since java 17 is out and it is LTS then Java 11 should be the minimum requirement. I can bring Java 8 support but if as you said, this project may used for millions developers, we need a fraction of them to contribute back, If that happens, I have no objection to keep Java 8 support for a while but it must be clear that it can't last forever and we need the industry to move forward.
> > Likely, 1.0 will be require Java 11 at minimum.
> 
> Well, thank you for your reply. I agree with that opinion to move forward, but please consider our request at the moment.

I do, but please help the project if you care about Java 8 support.
Alone, I can't do that

lburgazzoli:
> > @lburgazzoli I really also require java 8 support, I recently completed same things like that. I saw a lot of progress in this project, what's your plan for it ?
> 
> I did this pr only for those, who ask for java 8 support, but soon java 17 will be new LTS, so java 8 won't be long and only if @lburgazzoli decides to include java8 support

If you fix the conflict, I can merge it and bring back java 8 support for some time

lpcy:
> > > What I want to do is to support the latest LTS -1, so since java 17 is out and it is LTS then Java 11 should be the minimum requirement. I can bring Java 8 support but if as you said, this project may used for millions developers, we need a fraction of them to contribute back, If that happens, I have no objection to keep Java 8 support for a while but it must be clear that it can't last forever and we need the industry to move forward.
> > > Likely, 1.0 will be require Java 11 at minimum.
> > 
> > 
> > Well, thank you for your reply. I agree with that opinion to move forward, but please consider our request at the moment.
> 
> I do, but please help the project if you care about Java 8 support. Alone, I can't do that

OK，I'm pleasure to do my best.  I admire your work and believe that this project will be more  popular !

gradle.properties
sudoHackIn: Yes, i understand it very well and would like to help you to support this project. Ok, i'l try to create custom implementation without this plugin

jetcd-core/src/main/java/io/etcd/jetcd/impl/AuthCredential.java
```diff
@@ -85,7 +85,7 @@ private void authenticate(MetadataApplier applier) {
 
         if (!interceptorsChain.isEmpty()) {
             authFutureStub = authFutureStub.withInterceptors(
-                interceptorsChain.toArray(ClientInterceptor[]::new));
+                interceptorsChain.toArray(new ClientInterceptor[0]));
```
lburgazzoli, line 88: is this really needed ? afaik it should work on java 9 too

jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/EtcdContainer.java
```diff
@@ -65,7 +66,7 @@ public EtcdContainer withClusterToken(String clusterToken) {
 
     public EtcdContainer withAdditionalArgs(Collection<String> additionalArgs) {
         if (additionalArgs != null) {
-            this.additionalArgs = List.copyOf(additionalArgs);
+            this.additionalArgs = Collections.unmodifiableCollection(additionalArgs);
```
lburgazzoli, line 69: a ocpy is needed here

gradle/libs.versions.toml
```diff
@@ -20,7 +20,7 @@ errorprone = "2.10.0"
 vertx = "4.2.2"
 
 versionsPlugin = "0.39.0"
-versionsFilterPlugin = "0.1.10"
+versionsFilterPlugin = "0.1.4"
```
lburgazzoli, line 23: this needs to be either removed or replaced, otherwise it will be listed among the dependencies to upgrade

jetcd-core/src/main/java/io/etcd/jetcd/impl/AuthCredential.java
sudoHackIn: it works on java 9, but don't on java 8 this error - 
`error: incompatible types: Array is not a functional interface
                interceptorsChain.toArray(ClientInterceptor[]::new));`

lburgazzoli: ah ok

jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/EtcdContainer.java
sudoHackIn: `Collections.unmodifiableCollection(new ArrayList<>(additionalArgs));` will be ok?

lburgazzoli: yes

gradle/libs.versions.toml
sudoHackIn: i will do custom implementation and remove plugin

gradle/versions.gradle
```diff
@@ -15,11 +15,16 @@
...
 apply plugin: 'com.github.ben-manes.versions'
-apply plugin: 'se.ascp.gradle.gradle-versions-filter'
 
-versionsFilter {
-    exclusiveQualifiers = ['alpha', 'beta', 'rc', 'cr', 'm', 'ea']
-    log = false
+dependencyUpdates.configure {
+
+    def isNonStable = { String version ->
+        return ['alpha', 'beta', 'rc', 'cr', 'm', 'ea'].any { keyword -> version.toUpperCase().contains(keyword.toUpperCase())}
```
sudoHackIn, line 22: do we need more complex check with regex (as it was in version plugin)?
https://github.com/janderssonse/gradle-versions-filter-plugin/blob/c2682188f0f4fe5b36d72f5cb9561a9222ca4e50/src/main/kotlin/se/ascp/gradle/GradleVersionsFilterPlugin.kt#L135

lburgazzoli: I don't think so

lburgazzoli:
@sudoHackIn the PR looks good to me, do you mind updating also the readme to mention that Java 8 is supported ?

sudoHackIn:
> @sudoHackIn the PR looks good to me, do you mind updating also the readme to mention that Java 8 is supported ?

sure

lburgazzoli:
thx

PR #1050, opened 2022-03-14
Fix: use user-provided keep-alive timeouts as initial keepalive timeout

### LINKED ISSUE #1049: Use keep-alive-timeout from ClientBuilder instead of hardcoded FIRST_KEEPALIVE_TIMEOUT_MS
**Versions**
- jetcd: 0.5.10

**Describe the bug**

When network latency is high, the `KeepAlive` in `LeaseImpl` is removed by `deadLineExecutor` before first keep alive response from ETCD server, because it uses `FIRST_KEEPALIVE_TIMEOUT_MS=5000`

**To Reproduce**
Use byteman to inject 10 second delay to `LeaseImpl#processKeepAliveResponse`.

**Expected behavior**
It should use user-provided keep-alive timeout from `ClientBuilder` if present.

jasonstack:
Use user configurable keep-alive timeouts to avoid KeepAlive being
removed before receiving first response from server

Fixes #1049

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/ClientConnectionManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 The jetcd authors
+ * Copyright 2016-2022 The jetcd authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


lburgazzoli:
I don't have pan to backport at this stage

jasonstack:
@lburgazzoli we are currently using `0.5.8` in production, it'd be great to upgrade only patch version, instead of minor version. If you are concerning about behavior change in `0.5.x`, I can simplify the patch for `0.5.x` to make `FIRST_KEEPALIVE_TIMEOUT_MS=Integer.getInteger("jetcd.first_keep_alive_timeout_ms", 5000);` configurable.

jasonstack:
Got it

Issue #992, opened 2021-10-09
refreshToken failed


johnxiaohe:
My Etcd Server is Etcd Grpc-Proxy Cluster By Etcd Version 3.3.25

When Token is Invalid. Proxy Node response msg is ”etcdserver:invalid auth token“.
But response code is UNKNOWN(2), Not UNAUTHENTICATED(16). 
![企业微信截图_1633754973301](https://user-images.githubusercontent.com/37014780/136654589-e4631786-7a43-49d9-bedc-d3a84081fced.png)

lburgazzoli:
please add kore details, what is failing ?

johnxiaohe:
![image](https://user-images.githubusercontent.com/37014780/137616762-cf284967-e7b7-4e8e-9441-d3c208a99069.png)
```test
"C:\Program Files\Java\jdk1.8.0_261\bin\java.exe" -Dmaven.multiModuleProjectDirectory=D:\Workspace\JavaOpenSource\jetcd -Dmaven.home=E:\ToolApplication\Maven\apache-maven-3.6.3 -Dclassworlds.conf=E:\ToolApplication\Maven\apache-maven-3.6.3\bin\m2.conf "-Dmaven.ext.class.path=E:\IDEApplication\IDEA\IntelliJ IDEA 2020.2.4\plugins\maven\lib\maven-event-listener.jar" "-javaagent:E:\IDEApplication\IDEA\IntelliJ IDEA 2020.2.4\lib\idea_rt.jar=5034:E:\IDEApplication\IDEA\IntelliJ IDEA 2020.2.4\bin" -Dfile.encoding=UTF-8 -classpath E:\ToolApplication\Maven\apache-maven-3.6.3\boot\plexus-classworlds-2.6.0.jar;E:\ToolApplication\Maven\apache-maven-3.6.3\boot\plexus-classworlds.license org.codehaus.classworlds.Launcher -Didea.version=2020.2.4 org.xolstice.maven.plugins:protobuf-maven-plugin:0.6.1:compile
[INFO] Scanning for projects...
[INFO] ------------------------------------------------------------------------
[INFO] Detecting the operating system and CPU architecture
[INFO] ------------------------------------------------------------------------
[INFO] os.detected.name: windows
[INFO] os.detected.arch: x86_64
[INFO] os.detected.bitness: 64
[INFO] os.detected.version: 10.0
[INFO] os.detected.version.major: 10
[INFO] os.detected.version.minor: 0
[...]

PR #1059, opened 2022-04-01
invalid etcd token with Unkonwn status code

vinsia:
## Background
During using the etcd 3.4.16 and jetcd 0.5.10. I found that in a long running jetcd client. It will raise the Exception like `Caused by: io.etcd.jetcd.shaded.io.grpc.StatusRuntimeException: UNKNOWN: etcdserver: invalid auth token`.  The jetcd won't try to refresh the etcd token.
Also, I found the same issue: https://github.com/etcd-io/jetcd/issues/992. But he did not provide the fix. So in this PR, I just include the UNKONWN status code to refresh the etcd token.

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/support/Errors.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/support/Errors.java
@@ -20,6 +20,7 @@ import io.grpc.Status;
 
 public final class Errors {
     public static final String NO_LEADER_ERROR_MESSAGE = "etcdserver: no leader";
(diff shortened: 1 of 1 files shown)


vinsia:
@lburgazzoli PTAL, I'm trying to use the jetcd in the company to implement a long run client. But it will be eventually crashed and can not be recovered by himself.

lburgazzoli:
@vinsia any way to add a test for such scenario ?

lburgazzoli:
thx

yijianli:
@lburgazzoli do you have the release ETA? thanks

lburgazzoli:
hopefully this week

vinsia:
@lburgazzoli thanks

Issue #1057, opened 2022-03-29
Set daemon to true when creating background thread pool, and also set a proper name for the thread


Apache9:
**Is your feature request related to a problem? Please describe.**
In our company, some users do not close etcd client properly and non daemon thread will make their process hang there forever.
We can pass in an executorService when constructing etcd client through ClientBuilder, but in WatchImpl we will always create a non daemon thread pool. And what's more, the thread just uses the default name pattern so it is hard to identify what it is used for when checking the jstack output.

**Describe the solution you'd like**
Creating daemon thread by default, and give the thread a proper name which is easy to identify

**Describe alternatives you've considered**
Add a method in ClientBuilder to pass in a ExecutorService to be used by WatchImpl, so we can control it by our own.

**Additional context**
No.

yijianli:
We face the same issue, +1

lburgazzoli:
@Apache9 any chance you contribute a fix ?

Apache9:
Sure, let me provide a PR.

Apache9:
Opened #1062 

Do I need to sign something like an ICLA before contributing code?

Thanks.

lburgazzoli:
no, everything is fine

PR #1062, opened 2022-04-10
Set daemon to true when creating background thread pool, and also set…

Apache9:
… a proper name for the thread

Fix for #1057

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/impl/ClientConnectionManager.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/impl/ClientConnectionManager.java
@@ -20,6 +20,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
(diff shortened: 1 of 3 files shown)


jetcd-core/src/main/java/io/etcd/jetcd/impl/ClientConnectionManager.java
```diff
@@ -73,7 +74,15 @@
         this.credential = new AuthCredential(this);
 
         if (builder.executorService() == null) {
-            this.executorService = Executors.newCachedThreadPool();
+            final ThreadFactory backingThreadFactory = Executors.defaultThreadFactory();
```
vinsia, line 77: can you put this into https://github.com/etcd-io/jetcd/blob/main/jetcd-core/src/main/java/io/etcd/jetcd/support/Util.java to reuse?

Apache9: Let me give it a try.

Apache9: Done. PTAL?

vinsia (approved):
Thanks

PR #1065, opened 2022-04-11
Fix build instructions description in README.md

tenghuanhe:
Actually jetcd is built with `gradle` instead of `maven`


tenghuanhe:
Hi, @lburgazzoli could you please help have a look? Thank you.

PR #1068, opened 2022-04-24
Shadow the jetcd-all

vinsia:
### Background
This PR is going to provide a shadow `jetcd-core` package. The following dependencies will be shaded:
1. `io.grcp.*`
2. `io.netty.*`

The above two packages are often imported by other projects. Besides, shadow all the related SPI files. like `META-INF/io.grpc.NameResolverProvider`

### Testing
Have manually tested with this package, it works well. with conflict packages that mentioned above.


vinsia:
@lburgazzoli PTAL, thanks

vinsia:
@lburgazzoli ping

lburgazzoli (changes requested):
I would like to have a dedicated module instead of having a shaded jar part of jetcd-core

yijianli:
use a dedicated jetcd repo

yijianli:
@lburgazzoli PTAL

build.gradle
```diff
@@ -81,5 +81,7 @@ subprojects {
         theme 'mocha-parallel'
         showStandardStreams false
     }
+
+    tasks.withType(Javadoc).all { enabled = false }
```
lburgazzoli, line 85: what is this about ?

jetcd-all/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+dependencies {
+    api project(':jetcd-core')
+}
+
+shadowJar {
+    relocate 'io.grpc', 'io.etcd.jetcd.shaded.io.grpc'
+    relocate 'io.netty', 'io.etcd.jetcd.shaded.io.netty'
+    mergeServiceFiles()
+
+    archiveClassifier.set('')
```
lburgazzoli, line 31: what is this about ?

vinsia: by default, the shadow jar will add the classifier `all` to the end. so the package name will be `jetcd-all-0.7.1-SNAPSHOT-all.jar`. This command will try to remove the classifier. So the name will be `jetcd-all-0.7.1-SNAPSHOT.jar` eventually.

build.gradle
vinsia: removed, the java doc is not working in my side previously.
Test again, it should works now.

jetcd-all/build.gradle
vinsia: I thought since we are using a dedicated repo, we do not need the classifier now.

PR #1092, opened 2022-09-17
Add param of withBindVolumn for EtcdClusterExtension

liangyuanpeng:
I use remote docker for testcontainer to run unit test and i just got the error:

```
Caused by: com.github.dockerjava.api.exception.InternalServerErrorException: Status 500: {"message":"invalid volume specification: 'C:\\Users\\Administrator\\AppData\\Local\\Temp\\jetcd_test_etcd0_8631704060335352785:/data.etcd:rw,z'"}
```  

So i add param of bindVolumn for this case that do not need volumn.

### DIFF
--- a/jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/Etcd.java
+++ b/jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/Etcd.java
@@ -46,6 +46,7 @@ public final class Etcd {
         private boolean ssl = false;
         private List<String> additionalArgs;
         private Network network;
+        private boolean bindVolumn = true;
 
         public Builder withClusterName(String clusterName) {
             this.clusterName = clusterName;
@@ -87,6 +88,11 @@ public final class Etcd {
             return this;
         }
 
+        public Builder withBindVolumn(boolean bindVolumn) {
+            this.bindVolumn = bindVolumn;
+            return this;
+        }
+
         public EtcdCluster build() {
             return new EtcdClusterImpl(
                 image,
@@ -95,7 +101,8 @@ public final class Etcd {
                 nodes,
                 ssl,
                 additionalArgs,
-                network != null ? network : Network.newNetwork());
+                network != null ? network : Network.newNetwork(),
+                    bindVolumn);
         }
     }
 }
--- a/jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/EtcdClusterImpl.java
+++ b/jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/EtcdClusterImpl.java
@@ -43,7 +43,8 @@ public class EtcdClusterImpl implements EtcdCluster {
         int nodes,
         boolean ssl,
         Collection<String> additionalArgs,
-        Network network) {
+        Network network,
+        boolean bindVolumn) {
 
         this.clusterName = clusterName;
         this.network = network;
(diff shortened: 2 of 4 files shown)


lburgazzoli:
It would be nice to understand why the volume mount fails, ideally his should not be something you should be able to disable as it may be required by the jetcd container

liangyuanpeng:
> It would be nice to understand why the volume mount fails,

Run unit test on windows, and use remote docker container from linux, the voulme is `C:\\Users\\Administrator\\xxx:/data.etcd:rw,z`, this format can not use by docker container. So just got the error when container start.

>  ideally this should not be something you should be able to disable as it may be required by the jetcd container
Absolutely right, so still create volum for etcd test container by default,  User need to use `withBindVolumn(false)` for disable volumn when they want to this.

jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/Etcd.java
```diff
@@ -46,6 +46,7 @@ public static class Builder {
         private boolean ssl = false;
         private List<String> additionalArgs;
         private Network network;
+        private boolean bindVolumn = true;
```
lburgazzoli, line 49: - the default value is usually set in the constructor
- maybe use a better name that describe the intention i.e. bindDataDirectory

jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/EtcdContainer.java
```diff
@@ -203,7 +211,9 @@ public void start() {
     @Override
     public void close() {
         super.close();
-        deleteDataDirectory(dataDirectory);
+        if(bindVolumn) {
```
lburgazzoli, line 214: this check may not be needed as if the data directory os null, then the method won't do anything

jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/EtcdContainer.java
```diff
@@ -79,6 +80,11 @@ public EtcdContainer withClusterToken(String clusterToken) {
         this.clusterToken = clusterToken;
         return self();
     }
+    
+    public EtcdContainer withBindVolumn(boolean bindVolumn){
+        this.bindVolumn=bindVolumn;
```
lburgazzoli, line 85: probably a checkstyle violation, please reformat the code

jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/EtcdClusterImpl.java
```diff
@@ -57,6 +58,7 @@ public EtcdClusterImpl(
                     .withClusterToken(clusterName)
                     .withSll(ssl)
                     .withAdditionalArgs(additionalArgs)
+                        .withBindVolumn(bindVolumn)
```
lburgazzoli, line 61: must be formatted

lburgazzoli (changes requested):
some minor changes are needed

lburgazzoli:
@liangyuanpeng can you fix the findings ?

IlyasYOY:
@lburgazzoli is this PR up to date? 

I opened a #1123, I guess they are almost identical. 

I have all code formatter + tests there. This feature is very important to me. 

Thanks!

lburgazzoli:
> @lburgazzoli is this PR up to date?
> 
> I opened a #1123, I guess they are almost identical.
> 
> I have all code formatter + tests there. This feature is very important to me.
> 
> Thanks!

Merged, THX !

lburgazzoli:
Done in https://github.com/etcd-io/jetcd/pull/1123

PR #1098, opened 2022-10-29
feat: add isLearner in Member and MemberAddRequest

tsonglew:


jetcd-core/src/main/java/io/etcd/jetcd/Cluster.java
```diff
@@ -42,9 +42,10 @@ public interface Cluster extends CloseableClient {
      * add a new member into the cluster.
      *
      * @param  peerAddrs the peer addresses of the new member
+     * @param  isLearner whether the member is raft learner
      * @return           the response
      */
-    CompletableFuture<MemberAddResponse> addMember(List<URI> peerAddrs);
+    CompletableFuture<MemberAddResponse> addMember(List<URI> peerAddrs, boolean isLearner);
```
lburgazzoli, line 48: Is this mandatory or there is a default value ?

tsonglew: it's better to be mandatory in my opinion

lburgazzoli: do you have a link to the documentation on etcd ?

tsonglew: > do you have a link to the documentation on etcd ?

https://github.com/etcd-io/etcd/blob/c2378be1b552663ac6c5278148aefbeb281b7228/api/etcdserverpb/rpc.proto#L879

lburgazzoli: That's the proto definition, the documentation I was referring to is this one https://etcd.io/docs/v3.5/learning/design-learner/

jetcd-core/src/test/java/io/etcd/jetcd/impl/ClusterClientTest.java
```diff
@@ -73,7 +73,7 @@ public void testMemberManagement() throws ExecutionException, InterruptedExcepti
         final Client client = Client.builder().endpoints(n1.clientEndpoints()).build();
         final Cluster clusterClient = client.getClusterClient();
 
-        Member m2 = clusterClient.addMember(n2.peerEndpoints())
+        Member m2 = clusterClient.addMember(n2.peerEndpoints(), false)
```
lburgazzoli, line 76: we need a test for isLearner = true

jetcd-core/src/main/java/io/etcd/jetcd/Cluster.java
lburgazzoli: Since this is a breaking change, it would be nice to have a default method with the old behavior

lburgazzoli:
@tsonglew I added some comments

jetcd-core/src/main/java/io/etcd/jetcd/Cluster.java
tsonglew: Sorry, I didn't find any etcd doc related with `isLearner`, but I added another `addMember` method for the old behavior

jetcd-core/src/test/java/io/etcd/jetcd/impl/ClusterClientTest.java
tsonglew: Tests of `addMember(peerAddrs)`, `addMember(peerAddrs, true)`, `addMember(peerAddrs, false)` has been updated in the latest commit

jetcd-core/src/test/java/io/etcd/jetcd/impl/ClusterClientTest.java
```diff
@@ -90,4 +90,32 @@ public void testMemberManagement() throws ExecutionException, InterruptedExcepti
...
+        final Client client = Client.builder().endpoints(n1.clientEndpoints()).build();
+        final Cluster clusterClient = client.getClusterClient();
+
+        Member m2 = clusterClient.addMember(n2.peerEndpoints(), true)
+            .get(5, TimeUnit.SECONDS)
+            .getMember();
+
+        assertThat(m2).isNotNull();
+        assertThat(m2.isLearner()).isTrue();
+        assertThat(clusterClient.listMember().get().getMembers()).hasSize(2);
```
lburgazzoli, line 119: should we also check that one of the member is learner ?

jetcd-core/src/test/java/io/etcd/jetcd/impl/ClusterClientTest.java
```diff
@@ -90,4 +90,32 @@ public void testMemberManagement() throws ExecutionException, InterruptedExcepti
...
+        final Client client = Client.builder().endpoints(n1.clientEndpoints()).build();
+        final Cluster clusterClient = client.getClusterClient();
+
+        Member m2 = clusterClient.addMember(n2.peerEndpoints(), false)
+            .get(5, TimeUnit.SECONDS)
+            .getMember();
+
+        assertThat(m2).isNotNull();
+        assertThat(m2.isLearner()).isFalse();
+        assertThat(clusterClient.listMember().get().getMembers()).hasSize(2);
```
lburgazzoli, line 105: should we also check that none of the members is learner ?

lburgazzoli (approved):

PR #1103, opened 2022-11-07
Add handling for "etcdserver: revision of auth store is old"

darvay:
Error experienced by customers when Role changes are made to the ETCD Cluster.

We are seeing a use case when ETCD Role grants are updated we are getting an error "etcdserver: revision of auth store is old" on all JETCD clients right after the grant is applied. The only fix now is for the client to restart and reconnect.

ETCD Cluster config:

ETCD_AUTH_TOKEN: jwt,pub-key=/auth-token/authToken.pub,priv-key=/auth-token/authToken.key,sign-method=RS256,ttl=8h


lburgazzoli:
is there any way to add a test ?

darvay:
Thanks for the review @lburgazzoli , I wasn't sure how to trigger this error condition via unit tests, I see [WatchUnitTest.java](https://github.com/etcd-io/jetcd/blob/main/jetcd-core/src/test/java/io/etcd/jetcd/impl/WatchUnitTest.java#L422 ) does a check for that error message but didn't see a test for Errors.INVALID_AUTH_TOKEN_ERROR_MESSAGE . We have been relying on an integration test running Etcd 3.5.2 with multiple etcdctl role add/updates to try to trigger the error condition.

lburgazzoli:
Yeah we don have test for all the cases but if there is a way to test it, I would really appreciated given my very limited time to support the project

darvay:
@lburgazzoli let me know if this test will work

darvay:
@lburgazzoli hope all is well, did you have a chance to review the added test? Thanks

PR #1178, opened 2023-06-25
Add cache and schedule for ghaction

liangyuanpeng:
- Add github gradle cache for CI
- Add schedule to run ci every day

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -26,6 +26,7 @@ on:
       - main
   workflow_dispatch:
 
+
 permissions:
   contents: read
   
@@ -47,6 +48,7 @@ jobs:
         with:
           java-version: 11
           distribution: 'temurin'
+          cache: 'gradle'
       - name: Build Project
         run: |
           export TC_USER="$(id -u):$(id -g)"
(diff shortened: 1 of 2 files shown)


.github/workflows/build-pr.yml
```diff
@@ -25,6 +25,8 @@ on:
     branches:
       - main
   workflow_dispatch:
+  schedule:
+  - cron: "0 0 * * *"
```
lburgazzoli, line 29: what is the purpose for that ?

liangyuanpeng: Run CI with schedule  for discover flack tests early, does make sence?

lburgazzoli: this workflow only runs on PRs so I don't think it makes much sense

liangyuanpeng: Moved schedule to main workflow,and testing for run snapshot exclude schedule.

liangyuanpeng:
@lburgazzoli  PTAL,Thanks.

lburgazzoli:
@liangyuanpeng rebase required

liangyuanpeng:
@lburgazzoli  Done,Thanks.

PR #1180, opened 2023-06-30
[ghaction] martix CI for etcd version

liangyuanpeng:
Martix CI for etcd version,Improve client stability. 

We can add CI for etcd 3.6.x after it's released.

### DIFF
--- a/jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/Etcd.java
+++ b/jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/Etcd.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.UUID;
 
 import org.testcontainers.containers.Network;
+import org.testcontainers.shaded.com.google.common.base.Strings;
 
 public final class Etcd {
     public static final String CONTAINER_IMAGE = "gcr.io/etcd-development/etcd:v3.5.9";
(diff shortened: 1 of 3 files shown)


jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/Etcd.java
```diff
@@ -16,14 +16,10 @@
 
 package io.etcd.jetcd.launcher;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.UUID;
+import java.util.*;
```
lburgazzoli, line 19: please don't use star import

jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/Etcd.java
```diff
@@ -89,6 +85,10 @@ public Builder withNetwork(Network network) {
         }
 
         public EtcdCluster build() {
+            String imagePro = System.getenv("ETCD_IMAGE");
+            if (!Strings.isNullOrEmpty(imagePro)) {
+                image = imagePro;
```
lburgazzoli, line 90: this should be done in the constructor, otherwise it would override any call to `withImage`

lburgazzoli (changes requested):

jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/Etcd.java
liangyuanpeng: @lburgazzoli PTAL,Thanks.

jetcd-launcher/src/main/java/io/etcd/jetcd/launcher/Etcd.java
```diff
@@ -48,6 +50,13 @@ public static class Builder {
         private Network network;
         private boolean shouldMountDataDirectory = true;
 
+        public Builder(){
+            String image = System.getenv("ETCD_IMAGE");
+            if (!Strings.isNullOrEmpty(image)) {
+                withImage(image);
```
lburgazzoli, line 56: invoking a non final method from the constructor is usually a bad practice as could lead to unexpected results, better to set the instance variable directly.

lburgazzoli (approved):

liangyuanpeng:
Thanks for review :)

Issue #1188, opened 2023-07-11
The value of member id is out of bounds, and the actual value has exceeded the range of long.


tzfun:
**Versions**
- etcd: 3.5.0
- jetcd: 0.6.1
- java: 11

**Describe the bug**
The actual value is 18305956170087212562

However, the value obtained in the jetcd interface is -8170086329776577000

The same problem exists in the Leader id, which makes me wonder whether the long value in other places is also out of bounds?

lburgazzoli:
I think it is pretty much expected as Java does not have unsigned integer so unit64 is mapped to 'long' which if you print it can overflow but it is bit by bit correct.

You can use 'Long.toUnsignedString(...)' to get an unsigned string

See the protobuf reference guide to know more about type mapping: https://protobuf.dev/programming-guides/proto3/

Let me know if this answers your question/issue.

tzfun:
Yes, this is how I currently handle it. I think that the current long type should be replaced by a wider range of types in the interface definition, such as double, so that at least the caller can get readable data. 😅

lburgazzoli:
Well, it is already machine readable :) but yes, we can wrap it as a  BigInteger but if it is for the only purpose of having it nicely represented on screen, well then it can be easily done by the library consumer when/where needed.

If you have time, then you can provide a PR, however mind not to break backward compatibility.

tzfun:
Yes, I am now using BigInteger to do the conversion.

Ok, next I will pull a branch and try to add some interfaces, I will ensure backward compatibility.

PR #1190, opened 2023-07-12
Add some interfaces to support unsigned ID

tzfun:
Add som interfaces to support unsigned ID. Use BigInteger to convert unsigned long value.

feat #1188

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/Response.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/Response.java
@@ -16,6 +16,8 @@
 
 package io.etcd.jetcd;
 
+import java.math.BigInteger;
+
 /**
  * represents a generic Jetcd response.
  */
@@ -38,11 +40,31 @@ public interface Response {
         long getClusterId();
 
         /**
+         * Returns the cluster id.
+         * The value is actually an unsigned int64 value, which may exceed the range of long.
+         *
+         * @return unsigned cluster id
+         */
+        default BigInteger getUnsignedClusterId() {
+            return new BigInteger(Long.toUnsignedString(getClusterId()));
+        }
+
+        /**
          * Returns the member id
          *
          * @return the member id.
          */
         long getMemberId();
+
+        /**
+         * Returns the member id.
+         * The value is actually an unsigned int64 value, which may exceed the range of long.
+         *
+         * @return unsigned member id
+         */
+        default BigInteger getUnsignedMemberId() {
+            return new BigInteger(Long.toUnsignedString(getMemberId()));
+        }
 
         /**
          * Returns the revision id
--- a/jetcd-core/src/main/java/io/etcd/jetcd/cluster/Member.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/cluster/Member.java
@@ -16,6 +16,7 @@
 
 package io.etcd.jetcd.cluster;
 
+import java.math.BigInteger;
 import java.net.URI;
 import java.util.List;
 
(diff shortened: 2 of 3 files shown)


liangyuanpeng (approved):

jetcd-core/src/main/java/io/etcd/jetcd/Response.java
```diff
@@ -37,13 +39,33 @@ interface Header {
          */
         long getClusterId();
 
+        /**
+         * Returns the cluster id.
+         * The value is actually an unsigned int64 value, which may exceed the range of long.
```
lburgazzoli, line 44: it is an uint64, not an int64

jetcd-core/src/main/java/io/etcd/jetcd/Response.java
```diff
@@ -37,13 +39,33 @@ interface Header {
...
         /**
          * Returns the member id
          *
          * @return the member id.
          */
         long getMemberId();
 
+        /**
+         * Returns the member id.
+         * The value is actually an unsigned int64 value, which may exceed the range of long.
```
lburgazzoli, line 61: it is an uint64, not an int64

jetcd-core/src/main/java/io/etcd/jetcd/Response.java
```diff
@@ -37,13 +39,33 @@ interface Header {
...
         long getClusterId();
 
+        /**
+         * Returns the cluster id.
+         * The value is actually an unsigned int64 value, which may exceed the range of long.
+         *
+         * @return unsigned cluster id
+         */
+        default BigInteger getUnsignedClusterId() {
+            return new BigInteger(Long.toUnsignedString(getClusterId()));
```
lburgazzoli, line 49: Is there a way to avoid the intermediate string representation ? I'd like to avoid generating garbage if possible.

jetcd-core/src/main/java/io/etcd/jetcd/Response.java
```diff
@@ -37,13 +39,33 @@ interface Header {
...
         long getMemberId();
 
+        /**
+         * Returns the member id.
+         * The value is actually an unsigned int64 value, which may exceed the range of long.
+         *
+         * @return unsigned member id
+         */
+        default BigInteger getUnsignedMemberId() {
+            return new BigInteger(Long.toUnsignedString(getMemberId()));
```
lburgazzoli, line 66: Is there a way to avoid the intermediate string representation ? I'd like to avoid generating garbage if possible.

jetcd-core/src/main/java/io/etcd/jetcd/cluster/Member.java
```diff
@@ -38,6 +39,15 @@ public long getId() {
         return member.getID();
     }
 
+    /**
+     * Returns the member ID for this member.
+     *
+     * @return the unsigned id.
+     */
+    public BigInteger getUnsignedId() {
+        return new BigInteger(Long.toUnsignedString(getId()));
```
lburgazzoli, line 48: Is there a way to avoid the intermediate string representation ? I'd like to avoid generating garbage if possible.

lburgazzoli:
The PR looks ok, however now we have some more interface methods that are part of the API.

Given how trivial the solution is and the fact that this is mainly for representation purpose as from the machine POV, a long is perfectly fine, it is worth adding those methods ?

tzfun:
Yes, from the data level, these methods are unnecessary. The data in the machine is correct, and only needs to be converted when it is displayed to the client or user. 

I think these methods can be ignored because their application scenarios are relatively small.

But I think it can be explained in the code comments of these uint64 values, because from the perspective of the interface caller, they don’t know what the proto definition type is, and they don’t know that these data are unsigned int64, and they may think it is a bug.

lburgazzoli:
> But I think it can be explained in the code comments of these uint64 values, because from the perspective of the interface caller, they don’t know what the proto definition type is, and they don’t know that these data are unsigned int64, and they may think it is a bug.

Can you add such comments ?

tzfun:
Ok, I'll try to add it later

liangyuanpeng:
Increased user convenience I think is worth it.

PR #1208, opened 2023-08-24
replace usage of guava with standard java

### LINKED ISSUE #1207: Remove guava dependency
jetcd currently uses googles guava dependency, but in modern java this library is often no longer needed. Replacing usages of guava with standard jdk alternatives helps users of jetcd since they do not need an extra library which also might result in dependency conflicts.

Example:

`Preconditions.checkNotNull()` can be replaced with `java.util.Objects.requireNonNull()`

I will add a PR for this shortly

askoog:
Avoid adding extra third party dependency 'guava' by replacing methods with equivalent alternatives in standard jdk.

This helps users of etcd since every transitive third party dependency requires more code to be downloaded and might result in dependency conflicts.

fixes #1207

### DIFF
--- a/jetcd-common/build.gradle
+++ b/jetcd-common/build.gradle
@@ -17,6 +17,7 @@
 
 dependencies {
     api libs.slf4j
+    api libs.guava
     api libs.grpcCore
 
     testImplementation libs.bundles.testing
--- a/jetcd-common/build.gradle
+++ b/jetcd-common/build.gradle
@@ -17,7 +17,6 @@
 
 dependencies {
     api libs.slf4j
-    api libs.guava
     api libs.grpcCore
 
     testImplementation libs.bundles.testing
--- a/jetcd-common/src/main/java/io/etcd/jetcd/common/exception/EtcdException.java
+++ b/jetcd-common/src/main/java/io/etcd/jetcd/common/exception/EtcdException.java
@@ -16,7 +16,7 @@
 
 package io.etcd.jetcd.common.exception;
 
-import static com.google.common.base.Preconditions.checkNotNull;
+import java.util.Objects;
 
 /**
  * Base exception type for all exceptions produced by the etcd service.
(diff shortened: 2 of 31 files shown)


liangyuanpeng (approved):
make sense for me

lburgazzoli:
@askoog does this actually remove guava dependency ? I think this is still something that is transitively provided by grpc-java

liangyuanpeng:
I agree with reducing direct dependencies even though indirect dependencies may exist.

lburgazzoli:
the reason guava dependency is explicitly declared was to control it as unfortunately there were cases where guava was clashing anyway, but we can remove it and see how it goes. 

Unfortunately it does not prevent to break apps using jetcd as a library.

lburgazzoli:
@askoog @liangyuanpeng I'm not very convinced by removing the explicit dependency as guava is anyway leveraged by jetcd so it is a good practice to explicit defines dependencies the application depends on instead of rely on transitive dependencies as any change in the transitive dependency chain can break the app. 

So I'm fine to replace guava with standard java code but not to remove the dependency as long as jetcd still requires it.

WDYT ?

askoog:
wow, you guys are fast! Thanks for the quick response!

You are right @lburgazzoli, this does not actually remove the dependency since it's also transitively included. And I agree with you that you should declare the versions of your used transitive dependencies as well.

The problem we were trying to solve initially was that in some of our legacy applications we were bound to guava 20 which grpc seems to work fine with, but the [usage of the Streams class in jetcd](https://github.com/etcd-io/jetcd/pull/1208/files#diff-544b5cf3a80e507714122b11108532951c331d945a8aa2e4d89ae8ef1a31e604R141) was not compatible with such an old version. So by removing the usage of guava code directly but keeping the specified dependency version will work for us. I will revert the changes in the libs file so that the version is still specified there

lburgazzoli:
if you find out other guava bits that can be replaced by any standard java class, I'm more than happy.

In fact one of the reason we have some additional wrappers was exactly to avoid exposing guava as much as possible at well, that is a pain :)

lburgazzoli:
I don't see any build file with `libs.guava` as dependency, I think the modules where guava is required should explicit declare the guava dependency

lburgazzoli:
@askoog any update ?

lburgazzoli:
@askoog can you please rebase / fix conflicts ?

PR #1217, opened 2023-09-05
chore: remove trailing spaces and add newlines for non-Java files

johnshajiang:
Spotless plugin just takes charge of Java files, and no tool concerns non-Java files. So, some of non-Java files have many trailing spaces and miss newlines.


lburgazzoli (approved):

johnshajiang:
@lburgazzoli 
Thanks for approving this PR.
If no concern or problem, could you please merge this patch?

This commit changes dozens of files, then it possibly will conflict with future PRs.

liangyuanpeng:
@johnshajiang 

Just for confirm that this pr just working for once time , right?

It's not working if i create a new file and have not newlines.

johnshajiang:
@liangyuanpeng 
> Just for confirm that this pr just working for once time , right?
It's not working if i create a new file and have not newlines.

Yes.
The GitHub workflow should need one more check for this problem.

PR #1219, opened 2023-09-07
build: test with OpenJDK 17

johnshajiang:
Now that jetcd should support OpenJDK 11+, it would be better to test not only OpenJDK 11, but OpenJDK 17 as well.
For example, the issue fixed by #1216 can only be reproduced with OpenJDK 13+.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -42,6 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        java-version: [11, 17]
         etcd:
           - gcr.io/etcd-development/etcd:v3.5.9
           - gcr.io/etcd-development/etcd:v3.4.26
@@ -51,7 +52,7 @@ jobs:
       - name: Set Up Java
         uses: actions/setup-java@v3
         with:
-          java-version: 11
+          java-version: ${{ matrix.java-version }}
           distribution: 'temurin'
           cache: 'gradle'
       - name: Build Project
--- a/.github/workflows/build-main.yml
+++ b/.github/workflows/build-main.yml
@@ -34,6 +34,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        java-version: [11, 17]
         etcd:
           - gcr.io/etcd-development/etcd:v3.5.9
           - gcr.io/etcd-development/etcd:v3.4.26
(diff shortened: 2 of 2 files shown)


liangyuanpeng:
I think we can waiting for java 21.

.github/workflows/build-pr.yml
```diff
@@ -42,6 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        java-version: [11, 17]
```
liangyuanpeng, line 45: Also need to update https://github.com/etcd-io/jetcd/blob/main/.github/workflows/build-main.yml

johnshajiang:
@liangyuanpeng 
> I think we can waiting for java 21.

Not sure when Gradle can support JDK 21.
Please monitor https://github.com/gradle/gradle/issues/25574
In fact, it needs to upgrade Gradle to 8.3 for supporting JDK 20.

.github/workflows/build-pr.yml
johnshajiang: I did think about this point.
Frankly, I was not sure what are the purposes of `build-main.yml` and `build-tag.yml`.
So, I focus on PR only. And that is why the title is `test with OpenJDK 17 on PR`.

lburgazzoli: build tag was an attempt to automate the release process but it is not yet working

johnshajiang:
@lburgazzoli 
Could you please review this PR?
My following PR for issue #1221 will modify `build-pr.yml` as well.

Please feel free to give your suggestions or concerns.

liangyuanpeng (approved):

lburgazzoli (approved):

johnshajiang:
@lburgazzoli 
Thanks for approving and merging this PR!

In addition, I just have a suggestion.
It may be better to squash the multiple commits into a single commit when merge this PR.

lburgazzoli:
@johnshajiang I usually do ask the author to squash the commits as the author knows better. In this case I could probably have done it but that leaves to me the role of deciding the commit message and the description which may lead to remove some useful context the author wanted to preserve.

johnshajiang:
@lburgazzoli 
Please feel free to ask me to squash the commits after you finally approve my PR.

lburgazzoli:
> @lburgazzoli Please feel free to ask me to squash the commits after you finally approve my PR.

certainly, it was also very hard to get someone committed to do some good work so request to get reproducers, fix stuffs in PR and what not has always been a little bit not easy to achieve so as result I take the easy path and I fix things later on when I have time so any active contribution is more than welcome :)

Issue #1232, opened 2023-09-25
[Flake Test] testContainerShouldNotMountDirectory


liangyuanpeng:
```shell
5 tests completed, 2 failed
> Task :jetcd-launcher:test
  io.etcd.jetcd.launcher.test.EtcdClusterStartTest ✘ testContainerShouldNotMountDirectory() (7.2s)

    java.lang.IllegalStateException: Cluster failed to start
        at io.etcd.jetcd.launcher.test.EtcdClusterStartTest.testContainerShouldNotMountDirectory(EtcdClusterStartTest.java:60)
    Caused by: java.lang.RuntimeException: org.testcontainers.containers.ContainerLaunchException: Container startup failed for image gcr.io/etcd-development/etcd:v3.5.9

    Caused by: org.testcontainers.containers.ContainerLaunchException: Container startup failed for image gcr.io/etcd-development/etcd:v3.5.9

    Caused by: org.rnorth.ducttape.RetryCountExceededException: Retry limit hit with exception

    Caused by: org.testcontainers.containers.ContainerLaunchException: Could not create/start container

[...]

PR #1233, opened 2023-09-25
Skip exec chmod command when not mount data directory

liangyuanpeng:
Do not run command of chmod in container when have not mount data directory.

And hope this PR can resolved the falke test https://github.com/etcd-io/jetcd/issues/1232


liangyuanpeng:
@lburgazzoli  PTAL,Thanks, This is continuing to impact the main schedule workflow.

image

PR #1240, opened 2023-10-10
Etcd memory leak when using jetcd-core 0.7.6

### LINKED ISSUE #1236: Etcd memory leak when using jetcd-core 0.7.6
**Versions**
- etcd: 3.4.12
- jetcd: 0.7.6
- java: 17

**Describe the bug**
When using leases and keepAliveOnce function, there seems to be a memory leak inside of etcd, caused by jetcd-core version 0.7.6. This is easily reproducable by simply creating a key with a lease and using keepAliveOnce on that lease. To check if the problem exists, we can simply view etcd metrics (http://localhost:2379/metrics) and check the "go_goroutines" count. This count is rising out of control.  If I check the same code (same example project) using jetcd-core version 0.7.5 the problem is not present. Because of this I believe the problem to be caused by the following PR https://github.com/etcd-io/jetcd/pull/1165.

**To Reproduce**
Create a key with a lease and renew the lease periodically using keepAliveOnce. I have attached a simple example project which demonstrates this issue. Run the example project once with jetcd-core 0.7.6 and once with jetcd-core 0.7.5 and compare the outputs. The difference should be pretty big.
[example-project-jetcd-bug.zip](https://github.com/etcd-io/jetcd/files/12791865/example-project-jetcd-bug.zip)

**Expected behavior**
Etcd go_goroutines count should rise slightly (for this test count should be around 150 or less) and not as much as now (100k+ in attached test in 60 seconds)

**Additional context**
Because every go_routine uses a minimum of 8KB of memory, this can quickly spiral out of control and etcd instance consumes all available memory.

lburgazzoli:
Fixes #1236

### DIFF
--- a/jetcd-core/src/test/java/io/etcd/jetcd/impl/LeaseMemoryLeakTest.java
+++ b/jetcd-core/src/test/java/io/etcd/jetcd/impl/LeaseMemoryLeakTest.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright 2016-2021 The jetcd authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.etcd.jetcd.impl;
+
+import java.net.URI;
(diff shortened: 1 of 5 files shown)


jetcd-core/src/test/java/io/etcd/jetcd/impl/LeaseMemoryLeakTest.java
```diff
@@ -0,0 +1,108 @@
...
+                leaseClient.keepAliveOnce(leaseID).get(1, TimeUnit.SECONDS);
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+        int after = extractGoRoutinesCount(uri);
+        assertThat(after).isGreaterThan(0);
+
+        assertThat(after).isGreaterThan(before);
```
liangyuanpeng, line 87: Not sure if this makes this a flake test as there will be some fluctuations here.

PR #1337, opened 2024-03-12
Retry auth failures and require leader in ElectionImpl.

### LINKED ISSUE #1336: Election methods do not do auth retries, and do not require leader
**Versions**
- etcd: 3.5.12
- jetcd: 0.7.7
- java: 17

**Describe the bug**

1. The implementations of Election methods in `ElectionImpl.java` do not do auth retries: they do not use the `execute` method in `Impl`, which calls into the failsafe retry policy, instead they use the `completable` method, which doesn't.
The result for a follower that may be blocking in a campaign request for more than 5 minutes (the default time it takes for an auth token to expire) is, when promoted to leader, its campaign call will fail right away with "etcdserver: invalid auth token".

2. The implementation of Election methods (in `ElectionImpl.java`) does not set the metadata headers to require leader.  I believe it should, since participating or observing an election that becomes frozen client side due to a partitioned server defies the purpose of implementing fault tolerance on top of these primitives (also see Additional context section).

**To Reproduce**
Call `Election.campaign` and arrange for it to block after some other already elected leader, wait 5 minutes, kill the leader and instead of succeeding the campaign call will fail with "etcdserver: invalid auth token".

**Expected behavior**
The methods in Election should retry like every other API implementation method (eg, KV.get) does.
When the same situation with a follower that waits more than 5 minutes blocked in campaign and then becomes leader is simulated using etcdctl, the error shows as a warning but the call is automatically retried by the golang implementation:

```
# The DH_ETCD_DIR env var below sets up some role for authorization in a script we use to wrap etcdctl.
$ DH_ETCD_DIR=/etc/sysconfig/illumon.d/etcd/client/root etcdctl.sh elect /pepo/ 1
{"level":"warn","ts":"2024-03-11T17:33:03.864-0400","logger":"etcd-client","caller":"v3/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc000398c40/10.128.0.201:2379","attempt":0,"error":"rpc error: code = Unauthenticated desc = etcdserver: invalid auth token"}
/pepo//14d08e2f6108dd67
1
```

**Additional context**
Looking at the go client sources, it seems the go client (and by implication etcdctl) does not use the concurrency gRPC API, and instead directly implements the campaign and related calls as etcd operations (eg, campaign is implemented as a transaction and a watch, here: https://github.com/etcd-io/etcd/blob/e7b3bb6ccac840770f108ef9a0f013fa51b83256/client/v3/concurrency/election.go#L74) instead of making unary concurrency proto RPCs.  On one hand this lends evidence to the use of require leader for calling concurrency RPCs (since the get that is part of the transaction for campaign is a linearized read, etc).  On another hand, this is worrisome as it means the concurrency gRPC API is probably not as well tested; the behaviors are slightly different, eg, in the direct implementation in go client the blocking is implemented by waiting on responses on a watch, which is a streaming RPC, while in the RPC version of campaign is the unary campaign RPC itself that is blocking.  The server side implementation of concurrency RPC calls ends up calling the same go code in the client (eg, the transaction + watch for campaign, etc), so the low level implementation is the same, but the details of the RPC calls involved in clients (eg, go etcd versus jetcd) are different.

Also note that in go, adding "require leader" is an operation done on the gRPC context, and the campaign calls in go get that as a parameter, so it is possible for a caller to set that, and on second thought they should, since even if the transaction itself is going to require leader, the watcher does not by default I think.  In our case in jetcd there is no way to add require leader in the API.  My thought is that in both cases, go and jetcd, it should not be optional, given the use case for leader election methods, ie, implementing fault tolerance features.  I don't think optional makes sense in this context.  The same is probably true for lock. so perhaps we should do something there too.

jcferretti:
Fixes #1336

Also adds require leader in LockImpl; LockImpl was already using execute, so auth retries were happening there.

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/impl/ElectionImpl.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/impl/ElectionImpl.java
@@ -32,6 +32,7 @@ import io.etcd.jetcd.election.NoLeaderException;
 import io.etcd.jetcd.election.NotLeaderException;
 import io.etcd.jetcd.election.ProclaimResponse;
 import io.etcd.jetcd.election.ResignResponse;
+import io.etcd.jetcd.support.Errors;
 import io.etcd.jetcd.support.Util;
 import io.grpc.StatusRuntimeException;
 
@@ -51,6 +52,27 @@ final class ElectionImpl extends Impl implements Election {
         this.namespace = connectionManager.getNamespace();
     }
 
+    // Election operations are done in a context where a client is trying to implement
+    // some fault tolerance related use case; in that type of context, it makes sense to always
+    // apply require leader, since we don't want a client connected to a non-raft-leader server
+    // have an election method just go silent if the server the client happens to be connected to
+    // becomes partitioned from the actual raft-leader server in the etcd servers cluster:
+    // in that scenario and without required leader, an attempt to campaign could block forever
+    // not because some other client is already an election leader, but because the server the client
+    // is connected to is partitioned and can't tell.
+    // With require leader, in that case the call will fail and we give
+    // the client the ability to (a) know (b) retry on a different server.
+    // The retry on a different server should happen automatically if the connection manager is using
+    // a round robin strategy.
+    //
+    // Beware in the context of this election API, the word "leader" is overloaded.
+    // In the paragraph above when we say "raft-leader" we are talking about the etcd server that is a leader
+    // of the etcd servers cluster according to raft, we are not talking about the client that
+    // happens to be the leader of an election using the election API in this file.
+    private VertxElectionGrpc.ElectionVertxStub stubWithLeader() {
+        return Util.applyRequireLeader(true, stub);
+    }
+
     @Override
     public CompletableFuture<CampaignResponse> campaign(ByteSequence electionName, long leaseId, ByteSequence proposal) {
         requireNonNull(electionName, "election name should not be null");
@@ -62,10 +84,11 @@ final class ElectionImpl extends Impl implements Election {
             .setLease(leaseId)
             .build();
 
-        return completable(
-            stub.campaign(request),
-            CampaignResponse::new,
-            this::convertException);
+        return wrapConvertException(
+            execute(
+                () -> stubWithLeader().campaign(request),
+                CampaignResponse::new,
+                Errors::isRetryable));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


jcferretti:
Perhaps we should do the same for the lock API; my team is not currently using that so I did not think about it at first.  You can let me know if you'd like me to do the same changes for lock, I think it would make sense.

lburgazzoli:
@jcferretti whatever you can do to help the project would be more than welcome since my time, as you have noticed, is very very limited

jcferretti:
Understood, will do.  I will add to this same PR and change the title to reflect the scope.  I'll have something to you by tomorrow.

jcferretti:
Done, PTAL at your convenience.

jcferretti:
Tests pass locally for me:
```
cfs@tenten 23:40:08 ~/github/jetcd
$ ./gradlew test
Starting a Gradle Daemon (subsequent builds will be faster)
Java HotSpot(TM) 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended

> Task :jetcd-core:test


  io.etcd.jetcd.op.TxnTest ✔ testIfs()

[...]
  io.etcd.jetcd.impl.WatchResumeTest > testWatchOnPut(int) ✔ [5] 60 (1m 10s)

  135 passing (4m 10s)
  12 pending


Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/8.5/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD SUCCESSFUL in 4m 15s
25 actionable tasks: 3 executed, 22 up-to-date

```

The failure on the github PR checks is on `WatchTest.testWatchOnDeletion`:
```
[...]
  io.etcd.jetcd.impl.WatchTest > testWatchOnDelete(Client) ✔ [1] io.etcd.jetcd.impl.ClientImpl@5987e932
  io.etcd.jetcd.impl.WatchTest > testWatchOnDelete(Client) ✘ [2] io.etcd.jetcd.impl.ClientImpl@a518813 (30s)

    java.util.concurrent.TimeoutException: testWatchOnDelete(io.etcd.jetcd.Client) timed out after 30 seconds



WatchTest > testWatchOnDelete(Client) > [2] io.etcd.jetcd.impl.ClientImpl@a518813 FAILED
    java.util.concurrent.TimeoutException: testWatchOnDelete(io.etcd.jetcd.Client) timed out after 30 seconds
        at org.junit.jupiter.engine.extension.TimeoutExceptionFactory.create(TimeoutExceptionFactory.java:29)
        at org.junit.jupiter.engine.extension.SameThreadTimeoutInvocation.proceed(SameThreadTimeoutInvocation.java:58)
[...]
```

Not sure how the code in this PR could have broken that one.

lburgazzoli (approved):

lburgazzoli:
@jcferretti can you squash the commits ?

jcferretti:
Force-pushed to fix DCO (commit sign-off).

PR #1346, opened 2024-03-23
core: update errors retryable condition.

### LINKED ISSUE #1344: JWT token is not refreshed if etcd responds with "auth: revision in header is old"
**Versions**
- etcd: 3.5.1
- jetcd: 0.7.7
- java: 8

**Describe the bug**
When using jetcd with authentication enabled, if the cached JWT token becomes invalid because it's revision has become obsolete and etcd responds with "auth: revision in header is old" error, jetcd fails to refresh the token and instead reuses the same one until it expires normally.

**To Reproduce**
1. Configure jetcd to use JWT token authentication
2. Make a request to etcd to obtain a JWT token for the given credentials
3. Update the etcd auth database revision by adding a new role/user/permission etc....

**Expected behavior**
The instantiated jetcd client continues to work but instead on every subsequent request the error "io.grpc.StatusRuntimeException: UNKNOWN: auth: revision in header is old" is returned

**Additional context**
Seems like the auth token validity logic must be updated in **io.etcd.jetcd.support.Errors**:
`
public static boolean isInvalidTokenError(Status status) {
      return (status.getCode() == Code.UNAUTHENTICATED || status.getCode() == Code.UNKNOWN) && "etcdserver: invalid auth token".equals(status.getDescription());
   }
`
An extra check for  "auth: revision in header is old" should be added here. 

After I added this my tests worked fine.

Rmarian:
Retry request in case of auth store revision old error from server

Fixes #1344.

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/support/Errors.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/support/Errors.java
@@ -27,7 +27,8 @@ public final class Errors {
     }
 
     public static boolean isRetryable(Status status) {
-        return Status.UNAVAILABLE.getCode().equals(status.getCode()) || isInvalidTokenError(status) || isAuthStoreExpired(status);
+        return Status.UNAVAILABLE.getCode().equals(status.getCode()) || isInvalidTokenError(status)
+            || isAuthStoreExpired(status);
     }
 
     public static boolean isInvalidTokenError(Throwable e) {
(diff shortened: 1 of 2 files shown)


lburgazzoli:
@Rmarian you'd need to format the code

Rmarian:
@lburgazzoli - is there a gradle task to format them automatically?

lburgazzoli:
[./gradlew :jetcd-core:spotlessApply](url)

Rmarian:
Thanks! Will try to do it later in the evening. Now I am a bit busy with work...

Rmarian:
Hi @lburgazzoli ,

Sorry for the lat update. I formatted the files.

PR #1393, opened 2024-08-08
match retry policy with go etcd client

lavacat:


lburgazzoli:
@lavacat thx for the PR, as you have noticed, I do have a very limited amount of time to dedicate to the project, so this PR is much appreciated

jcferretti:
I just found this PR after https://github.com/etcd-io/jetcd/pull/1445 had already been merged.

It would have helped if this PR had any linking information to the jepsen report and its dicussion in etcd github issues (that I assumed motivated it); as it was created, there was no context or motivation mentioned.

As I mentioned in the discussion thread in https://github.com/etcd-io/etcd/issues/18424#issuecomment-2661580643, I don't believe the conditions

```
return desc.equals("there is no address available") || desc.equals("there is no connection available");
```

can happen (I don't think the expression in that return can be true), since I do not think UNAVAILABLE is being generated client side here, but actually being sent by the server.

lavacat:
@vorburger sorry about slow reply. I'll try to rebase, give me some time

jcferretti:
I believe there is no point in this PR as https://github.com/etcd-io/jetcd/pull/1445 is already merged and it addresses the same root problem.  There is also some confusion about the whole issue in terms of what was happening in jetcd per jepsen and the different ways the jetcd and go clients work.  It would make sense to try to take a few steps back and understand what has happened before here, both in terms of the understanding around the original jepsen report and what was happening in terms of jetcd behavior and the changes that have been made since then.

lavacat:
Sorry, I got busy and didn't followup last year back when it was still fresh on my mind.

@jcferretti your PR looks good, but imo it will be better to follow the same naming convention as go client - `isSafeRetryMutableRPC` and `isSafeRetryImmutableRPC` for convenience of majority of contributors that are only familiar with go client.

Also, I wouldn't expose autoRetry option, since it's not exposed in go-client. Otherwise Jepsen can run with autoRetry = true and claim that etcd is broken.
The problem with Jepsen tests is that majority of etcd maintainers use go client and it requires non-trivial amount of time to setup jepsen/figure out jetcd differences.

Correct me if I'm wrong but your main concern was about 
```
return desc == "there is no address available" || desc == "there is no connection available" 
```

You are correct, those messages never happen. I've copied from go client for completeness. No need to have them in jetcd

jcferretti:
> Sorry, I got busy and didn't followup last year back when it was still fresh on my mind.
No worries, and thanks for taking the time now.

> @jcferretti your PR looks good, but imo it will be better to follow the same naming convention as go client - `isSafeRetryMutableRPC` and `isSafeRetryImmutableRPC` for convenience of majority of contributors that are only familiar with go client.
I don't see a problem with that, but I there is an implicit assumption in your statement that I don't think is true: that people that make contributions on the etcd client go side also frequently make contributions in jetcd / etcd client java side.  My company and my team has been using etcd with a java system via jetcd for many years, and we have been paying attention to this repo.  This is the first time I see some attention from the main etcd repo people on jetcd, which I am going to assume was motivated by the visibility of jepsen and the long winded discussion around their report.

> 
> Also, I wouldn't expose autoRetry option, since it's not exposed in go-client.
The option is important for giving users of jetcd a escape hatch that allows them to continue to run their systems in the ways they were running them before.  We are switching auto-retries off by default for the non safe cases (writes), which is a non-trivial change in behavior that can impact people.  They will need to audit their codebases and figure out some decision logic on their application level side for when, and if, to auto retry.  That will take time.   Giving them an option gives them breathing space to do that.

> Otherwise Jepsen can run with autoRetry = true and claim that etcd is broken.
autoRetry true is not the default.  There are many options that can be set incorrectly and break etcd or jetcd, I don't see how this is different.

> The problem with Jepsen tests is that majority of etcd maintainers use go client and it requires non-trivial amount of time to setup jepsen/figure out jetcd differences.
That is a problem, yes.  It looks to me like the collective etcd/jetcd community does care about jepsen tests given their visibility, and also given that they do help and have helped in the past to find issues.  The jepsen code is closer to the java ecosystem so is easier for jepsen to use a java based client library.  It is possible to convert those tests to interface with a go client instead (jepsen has tested systems without a java client, so it is possible).  The question is who will invest the time in doing something like that.  The argument from jepsen which they made explicit in the closing of their report text in their website is that jetcd is the official java client library for etcd.  So from a rational argument perspective I think you (etcd main project) have three options if you want to change this situation:

* Stop endorsing jetcd as an official client.  This would require explicitly declaring jetcd as a third party work in progress that is known to have issues.  Seems unfair, but is an option.
* Invest in moving the jepsen tests from jetcd to go client
* Invest in making jetcd closer to go client (I recognize that your original PR was an attempt in this direction)

You guys are short on effort and have plenty of other stuff to do, so all the "invest in" above is painful.  But then again, so is everybody (short on effort).

> 
> Correct me if I'm wrong but your main concern was about
> 
> ```
> return desc == "there is no address available" || desc == "there is no connection available" 
> ```
> 
> You are correct, those messages never happen. I've copied from go client for completeness. No need to have them in jetcd

Thanks.

PR #1445, opened 2025-02-16
Only retry idempotent operations by default. Fixes #1444

### LINKED ISSUE #1444: jetcd should not do gRPC retries for non-idempotent operations on gRPC UNAVAILABLE
**Versions**
- etcd: N/A
- jetcd: 0.8.3, also impacts earlier versions
- java: N/A

**Describe the bug**
jensen report https://jepsen.io/analyses/jetcd-0.8.2 explains issues that can occur due to automated retries of non-idempotent operations.  The issue has been discussed at length in etcd github issues, and some of the conclusions of those discussions captured for the go etcd client here: https://github.com/etcd-io/etcd/issues/18424 (the jensen report references the other etcd issues related)

**To Reproduce**
Refer to previous paragraph

**Expected behavior**
Non-idempotent operations like writes and transactions should not retry when it is not known at the client if the server may have performed the operation.  Eg, token expired is safe to retry, a general gRPC UNAVAILABLE error is not

**Additional context**
See the description for UNAVAILABLE in the gRPC documentation https://grpc.github.io/grpc/core/md_doc_statuscodes.html stating "Note that it is not always safe to retry non-idempotent operations"

**Proposed change**
I propose we change the implementation of `isRetryable` and remove `Status.UNAVAILABLE` from the or (`||`) expression.
https://github.com/etcd-io/jetcd/blob/a240dc6a74c82b96a4aab898a181bf35283dd386/jetcd-core/src/main/java/io/etcd/jetcd/support/Errors.java#L30
It would be possible to do better by either (a) keeping careful track of what operations is being performed (eg, we could retry GETs safely), or (b) etcd providing some other more detailed error response mechanism aside from defaulting to UNAVAILABLE in many different situations.  But those changes would be more complex and thus more risky.  This smaller change would prevent from retrying in cases where is not obviously safe with the tradeoff of increasing the burden on users to do application level retries in cases where we were doing them before.  That tradeoff sounds fair to me for correctness.

jcferretti:
See description on https://github.com/etcd-io/jetcd/issues/1444

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/KV.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/KV.java
@@ -26,6 +26,7 @@ import io.etcd.jetcd.options.CompactOption;
 import io.etcd.jetcd.options.DeleteOption;
 import io.etcd.jetcd.options.GetOption;
 import io.etcd.jetcd.options.PutOption;
+import io.etcd.jetcd.options.TxnOption;
 import io.etcd.jetcd.support.CloseableClient;
 
 /**
@@ -115,4 +116,12 @@ public interface KV extends CloseableClient {
      * @return a Txn
      */
     Txn txn();
+
+    /**
+     * creates a transaction.
+     *
+     * @param  option TxnOption
+     * @return        a Txn
+     */
+    Txn txn(TxnOption option);
 }
--- a/jetcd-core/src/main/java/io/etcd/jetcd/impl/ElectionImpl.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/impl/ElectionImpl.java
@@ -88,7 +88,7 @@ final class ElectionImpl extends Impl implements Election {
             execute(
                 () -> stubWithLeader().campaign(request),
                 CampaignResponse::new,
-                Errors::isRetryable));
+                Errors::isRetryableForNoSafeRedoOp));
     }
 
     @Override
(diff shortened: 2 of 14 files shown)


jcferretti:
Note the PR is also renaming a test case,
    `public void testAuthErrorIsNotRetryable()`
to
    `public void testAuthErrorIsRetryable()`
in trying to correctly reflect what the test is actually doing (the test was, and still is, checking that the error is retryable)

lburgazzoli:
@jcferretti thx a lot for the PR, wonder if the `autoRetry` flag could be added to the related `*Option` object, instead of an additional method argument.

jcferretti:
I used the chance to add a few more javadoc comments to other options, eg,
```
    /**
     * Only populate results for keys that match a
     * minimum value for a created revision.
     *
     * Note this filter does not affect the Count field in GetResponse.
     * Count always counts the number of keys matched on a range, independent of other filters.
     *
     * @return minimum created revision to match, or zero for any.
     */
    public long getMinCreateRevision() {
        return this.minCreateRevision;
    }
```

The semantics of count not matching requested filters bit me in the past: https://github.com/etcd-io/etcd/issues/18267
I am trying to help prevent it biting others in the future.

lburgazzoli (approved):

jcferretti:
@lburgazzoli thank you very much for the quick turnaround.  Any chance there will be a new minor release including this in the medium term?  My team would like to pick it up for our system once available.  Not immediately urgent for us, but since it impacts reliability, it would be nice to have some understanding/visibility for when.

lburgazzoli:
hope before of end of the week

lburgazzoli:
@jcferretti jetcd-0.8.5 has been released

jcferretti:
Thanks!

Issue #1486, opened 2025-06-30
vorburger@ continued Org Membership confirmation


vorburger:
https://github.com/kubernetes/org/issues/5665 prompted me to review my past involvement here (I kinda forgot!).

If I can help, I'm happy to make a new effort to occasionally review PRs and such.

I've (just now) adjusted my GitHub notification settings accordingly.

If I correctly understand the aforementioned issue, a current maintainer has to somehow "vouch" for me here, @Priyankasaggu11929 is that how this works?

@lburgazzoli @fanminshi 👋 do you still want me around? ☺️

PR #1487, opened 2025-06-30
chore: Add vorburger to OWNERS

vorburger:
Fixes https://github.com/etcd-io/jetcd/issues/1486,

for https://github.com/kubernetes/org/issues/5665.

@lburgazzoli OK ?

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -5,3 +5,4 @@ approvers:
   - lburgazzoli     # Luca Burgazzoli <lburgazzoli@gmail.com>
   - xiang90         # Xiang Li <xiang.li@coreos.com>
   - heyitsanthony   # Anthony Romano <anthony.romano@coreos.com>
+  - vorburger       # Michael Vorburger <mike@vorburger.ch>
--- a/OWNERS
+++ b/OWNERS
@@ -3,6 +3,4 @@
 approvers:
   - fanminshi       # Fanmin Shi <fanmin.shi@coreos.com>
   - lburgazzoli     # Luca Burgazzoli <lburgazzoli@gmail.com>
-  - xiang90         # Xiang Li <xiang.li@coreos.com>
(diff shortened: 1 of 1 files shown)


jberkus:
This file needs to be brought up to date with the central info about the subproject.  A SIG chair should do it:

@jmhbnz @siyuanfoundation

OWNERS
```diff
@@ -5,3 +5,4 @@ approvers:
   - lburgazzoli     # Luca Burgazzoli <lburgazzoli@gmail.com>
   - xiang90         # Xiang Li <xiang.li@coreos.com>
   - heyitsanthony   # Anthony Romano <anthony.romano@coreos.com>
+  - vorburger       # Michael Vorburger <mike@vorburger.ch>
```
jmhbnz, line 8: Hey team - We need to make sure https://github.com/kubernetes/org/blob/main/config/etcd-io/sig-etcd/teams.yaml#L79-L84 aligns with this file and vice versa.

Looks like it's time to refresh/update both.

Reviewing recent commits and https://github.com/etcd-io/jetcd/graphs/contributors it looks to me like this file as well as the counterpart in `k/org` should be updated to list just @lburgazzoli and @vorburger as the active OWNERS.

Do folks agree with that?  If so we can update this pr to reflect and then update `k/org` following.

vorburger: @jmhbnz and @jberkus I'm sure it's up to me to remove @fanminshi (who IS on the aforementioned file, and [the 2nd top contributor](https://github.com/etcd-io/jetcd/graphs/contributors); I'd suggest what we definitely keep at least him here as 3rd contrib) or @xiang90 or @heyitsanthony (who are not). Folks, seeing that you have apparently never actually contributed to this repo, I've taken the liberty of removing you.

I've fixed the DCO with (stupid) `git commit --amend -s` as well.

@jmhbnz @jberkus @lburgazzoli merge this as-is now?

PR #1502, opened 2025-07-28
feat: add fields - raftAppliedIndex , errorsList, dbSizeInUse and isLearner in StatusResponse

mahanteshwar-git:
## What this PR does

This PR adds support for retrieving additional fields in StatusResponse
- raftAppliedIndex
-  errorsList
- dbSizeInUse
- isLearner

## Why this is useful

AppliedRaftIndex in comparison with RaftIndex can be used to determine if the PromoteMember operation could be retried
isLearner call in StatusResponse removes the need for another call to MemberList to determine if the member is a learner

## Implementation Details

- Added a fields to the class: StatusResponse
- Updated `testMemberStatus` to verify that the new fields are indeed present and populated

## Related Issues / Discussions

This change was based on discussion in #1499

---

## Checklist

- [x] Code compiles successfully
- [x] Tests added / updated
- [x] Documentation updated (if needed)
- [x] Ready for review

---

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/maintenance/StatusResponse.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/maintenance/StatusResponse.java
@@ -16,6 +16,8 @@
 
 package io.etcd.jetcd.maintenance;
 
+import java.util.List;
+
 import io.etcd.jetcd.Maintenance;
 import io.etcd.jetcd.impl.AbstractResponse;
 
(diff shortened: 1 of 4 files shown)


jetcd-core/src/main/java/io/etcd/jetcd/maintenance/StatusResponse.java
```diff
@@ -73,4 +75,40 @@ public long getRaftIndex() {
...
+    public long getDbSizeInUse() {
+        return getResponse().getDbSizeInUse();
+    }
+
+    /**
+     * Returns if the member is a raft learner.
+     *
+     * @return if the member is a raft learner
+     */
+    public boolean getIsLearner() {
```
vorburger, line 111: Let's rename `getIsLearner()` to `isLearner()` for consistency; both in general but also e.g. specifically with the existing [`Member.isLearner()`](https://github.com/etcd-io/jetcd/blob/9b4c9dd2ce8a78bf9cd99e88b88af139e422ba87/jetcd-core/src/main/java/io/etcd/jetcd/cluster/Member.java#L74)?

```suggestion
    public boolean isLearner() {
```

vorburger (changes requested):

PR #1518, opened 2025-09-10
feat(watch): add Watch.Watcher#cancel to cancel a single watcher

### LINKED ISSUE #1510: ServerWatchStream can't be closed when client closed watcher
**Versions**
- etcd: 3.5.17
- jetcd: 0.8.3
- java: JDK21

**Describe the bug**
After multiple calls to add watcher and close watcher, the watcher streams inside etcd continue to increase without being released, resulting in a leak.

**To Reproduce**
while(1) {
Watch.Watcher watcher = etcdClient.getWatchClient().watch(....);
TimeUnit.SECONDS.sleep(5);
watcher.close();
}

**Expected behavior**
watcher.close() will close the watcher stream and watcher routine

**Additional context**
-----
func (sws *serverWatchStream) recvLoop() error {
	for {
		req, err := sws.gRPCStream.Recv()
		if errors.Is(err, io.EOF) { ------->client send end-stream
			return nil ---->return nil
		}
-----
	go func() {
		if rerr := sws.recvLoop(); rerr != nil { ------->EOF return nil, so cannot be true
			if isClientCtxErr(stream.Context().Err(), rerr) {
				sws.lg.Debug("failed to receive watch request from gRPC stream", zap.Error(rerr))
			} else {
				sws.lg.Warn("failed to receive watch request from gRPC stream", zap.Error(rerr))
				streamFailures.WithLabelValues("receive", "watch").Inc()
			}
			errc <- rerr ------->this will not happen
		}
	}()

---
	select {
	case err = <-errc:
		if errors.Is(err, context.Canceled) {
			err = rpctypes.ErrGRPCWatchCanceled
		}
		close(sws.ctrlStream)
	case <-stream.Context().Done():
		err = stream.Context().Err()
		if errors.Is(err, context.Canceled) {
			err = rpctypes.ErrGRPCWatchCanceled
		}
	}
select will be blocked, so  sws.close() will never executed

chaofengliu-okg:
scripts: add Watch.Watcher#cancel to cancel a single watcher

Currently, closing a watcher is only possible via Watcher#close, which merely sends an HTTP/2 END_STREAM frame. However, etcd does not terminate the underlying goroutines and watcher stream in response, which can lead to memory leaks. By adding a cancel function that directly sends an HTTP/2 RST_STREAM frame, etcd can properly shut down the goroutines and the watcher stream.

We can reproduce the issue with the following code：
        for () {
            Watch.Watcher watcher = watchClient.watch(
                    key,
                    (WatchResponse resp) -> {
                        System.out.println("Received event: " + resp);
                    },
                    (Throwable t) -> {
                        System.out.println("Error: " + t);
                    },
                    () -> {
                        System.out.println("Completed");
                    }
            );
            //watcher.cancel();
            watcher.close();
            System.in.read();
        }
After adding the cancel API, replacing watcher.close() with watcher.cancel() allows the etcd server to properly release the stream and goroutines.

Fixes #1510


chaofengliu-okg:
I've changed the modification status from 'cancel' to 'close'. Please review again.

lburgazzoli:
@chaofengliu-okg is there any way to add a test for this case ?

chaofengliu-okg:
​​Thank you for merging the PR. I would like to ask about the schedule for the next release.​

PR #1558, opened 2026-01-21
WIP: Support blocking calls in client response handlers

cbornet:


cbornet:
This doesn't work as the current vertx gRPC impl is just a wrapper around grpc-java and `StreamObserverReadStream::pause` is a no-op.
Handling back-pressure would require doing it at the grpc-java `ClientCallStreamObserver`/`ClientResponseObserver` level and is incompatible with using the stubs 😞

Issue #1372, opened 2024-06-18
Switch to vertx-grpc-client


lburgazzoli:
As today, the vertx-grpc library is still based on grpc-java which has a number of issues and given the time I have to maintain the project, it make maintenance harder hence, given the project is already vert.x based, wet should switch to the [vert.x grpc client](https://vertx.io/docs/vertx-grpc/java/#_vert_x_grpc_client)
 
- https://github.com/etcd-io/jetcd/issues/1370
- https://vertx.io/docs/vertx-grpc/java/#_vert_x_grpc_client
- https://github.com/eclipse-vertx/vertx-service-resolver
- https://github.com/eclipse-vertx/vertx-grpc/issues/100
- https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/examples/HTTPExamples.java#L1405
- https://github.com/eclipse-vertx/vertx-grpc/blob/main/vertx-grpc-client/src/main/java/io/vertx/grpc/client/GrpcClient.java#L95

Issue #1285, opened 2023-12-13
not able to read any other etcd key while processing a etcd watch event. its blocked indefinitely


ankit2610:
**Versions**
- etcd: v3.4.28
- jetcd: 0.7.6
- java: 17

**Describe the bug**
I have some keys in etcd like **key1/123** and **key2/456**.

I have created a watcher on delete of a key prefixed with  key1.
when I process a key1 delete event, I am trying to get all the keys prefixed with key2, and its blocked here.

**To Reproduce**
Create a watch for one key and while processing the watch event , read another key. It will hang indefinitely.
**Expected behavior**
it should be able to read etcd keys irrespective of the context from where it is called.

**Additional context**

lburgazzoli:
https://github.com/etcd-io/jetcd/issues/1089

ankit2610:
sure . thanks for the update

lburgazzoli:
@ankit2610 can this be closed ?

Issue #1089, opened 2022-08-18
Thread blocked on get key inside a watcher


q12321q:
**Versions**
- etcd: 3.5.1
- jetcd: 0.7.3
- java: 11.0.8

**Describe the bug**
I don't know if it's a bug or a miss usage of the jetcd sdk but since we migrate from jetcd version `0.5.7` to `0.7.3` when we do a get key within a watcher notification we get the exception `io.vertx.core.VertxException: Thread blocked` and indeed the get key is stuck.

**To Reproduce**
Here how I reproduce:
```java
    @Test
    public void testWatcherAndGet() throws Exception {

            final Client client = Client.builder().endpoints("http://localhost:32784").build();
            final ByteSequence keyToGet = ByteSequence.from("/keytoGet", StandardCharsets.UTF_8);
            final ByteSequence keyToWatch = ByteSequence.from("/keytoWatch", StandardCharsets.UTF_8);
            final ByteSequence keyValueToWatch = ByteSequence.from("keyValuetoWatch", StandardCharsets.UTF_8);

            CountDownLatch countDownLatch = new CountDownLatch(1);
            client.getWatchClient().watch(keyToWatch, WatchOption.DEFAULT, new Watch.Listener() {

                @Override
                public void onNext(WatchResponse watchResponse) {
                    try {
                        client.getKVClient().get(keyToGet).get(); // stuck there with 0.7.3 but ok with 0.5.7
                        countDownLatch.countDown();
                    } catch (Exception e) {
                        e.printStackTrace();
                    }
                }

                @Override
                public void onError(Throwable throwable) { }
                @Override
                public void onCompleted() { }
            });

            client.getKVClient().put(keyToWatch, keyValueToWatch).get();
            countDownLatch.await();
    }
```

Here is the full exception:
```
2022-08-18T16:36:31.680Z dev WARN  vertx-blocked-thread-checker i.v.c.i.BlockedThreadChecker Thread Thread[vert.x-eventloop-thread-0,5,main] has been blocked for 2413 ms, time limit is 2000 ms
2022-08-18T16:36:32.681Z dev WARN  vertx-blocked-thread-checker i.v.c.i.BlockedThreadChecker Thread Thread[vert.x-eventloop-thread-0,5,main] has been blocked for 3415 ms, time limit is 2000 ms
[...]

PR #1559, opened 2026-01-22
Use NettyChannelBuilder to enable blocking calls in callbacks

cbornet:
It's currently not possible to do blocking calls in listener handlers while preserving back-pressure as that blocks the vertx event loop (see https://github.com/etcd-io/jetcd/issues/1089 https://github.com/etcd-io/jetcd/issues/1285).
Instead of `VertxChannelBuilder`, we can use `NettyChannelBuilder` delegates `onNext`/`onError`/`onCompleted` to a thread-pool with ordering guarantees.
It is OK to do it as the vertx-grpc stubs are just wrappers around the grpc-java ones and don't actually use the vertx loop. So they are compatible with any `ManagedChannelBuilder` (eg. in unit tests `InProcessChannelBuilder` is used).

Alternatives: 
* Rejected: pass responses to an ordered blocking call executor. This solves the blocking issue but breaks the back-pressure chain (see #1558 )
* Future work: use the vertx-grpc-client (https://github.com/etcd-io/jetcd/issues/1372)

### DIFF
--- a/jetcd-core/src/main/java/io/etcd/jetcd/ClientBuilder.java
+++ b/jetcd-core/src/main/java/io/etcd/jetcd/ClientBuilder.java
@@ -70,6 +70,7 @@ public final class ClientBuilder implements Cloneable {
     private Duration connectTimeout;
     private boolean waitForReady = true;
     private Vertx vertx;
+    private boolean useVertx = false;
 
     ClientBuilder() {
     }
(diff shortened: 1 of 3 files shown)


cbornet:
Hi @lburgazzoli, would you have time to review ?

lburgazzoli:
hope by the end of the week, please ping me again in case I won't

cbornet:
> hope by the end of the week, please ping me again in case I won't

Hi @lburgazzoli, gentle ping.

cbornet:
@lburgazzoli , gentle ping ?

lburgazzoli:
hey sorry, I promise I'll look into this before EOW, sorry again

lburgazzoli (approved):

lburgazzoli:
@cbornet merged it (sorry for the long delay), 

Can you also look at:
- https://github.com/etcd-io/jetcd/issues/1372 
- https://github.com/lburgazzoli/etcd-io-jetcd/tree/vertx-client

cbornet:
Awesome, thanks!

cbornet:
> Can you also look at:
> https://github.com/etcd-io/jetcd/issues/1372

Yes, I already proposed a way to support blocking calls in listener handlers when using Vert.x : https://github.com/lburgazzoli/etcd-io-jetcd/pull/1

cbornet:
@lburgazzoli would it be possible to get a release with the fix ? Thanks in advance.

lburgazzoli:
I'll try to do a release over the week end

